PR #3, opened 2024-01-31
Frame and Kernel Allocator

mtoohey31:


src/main.rs
```diff
@@ -30,41 +33,62 @@ _start:
 
 #[allow(dead_code)]
 extern "C" fn start(magic: usize, multiboot2_info: *mut Info) -> ! {
```
DerekCresswell, line 35: This is general, and not specific to these changes, but I think one of the things that greatly helps PintOS code readability was that their main function was as simple as possible. Really just a set of calls to other initialization code. I think we should abstract main as such where possible.

src/mem/buddy_allocator.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        *region.as_ptr().cast::<State>() = State::Split;
+        *right.as_ptr().cast::<State>() = State::Free;
+        Self::divide_free(left, target_size)
+    }
+
+    unsafe fn deallocate_rec(region: NonNull<[u8]>, ptr: NonNull<u8>) -> bool {
+        match *region.as_ptr().cast::<State>() {
+            // We should never be searching for the region of a ptr in a region
```
DerekCresswell, line 87: This might be just some of my inexperience with rust showing, but the code here looks dense. I worry that if this is any part of student facing code they would have a hard time understanding it.
I think your comments are good, but you pack a lot into each line. Maybe that's just things like `region.as_ptr...` cluttering it up. Can we not just make that a variable and keep it as such?

DerekCresswell: The result of the cast*

DerekCresswell (commented):
I like the overall shape of the code. I think it can, funnily enough, be loosened up so as to be more accessible to students with little rust experience.

src/main.rs
mtoohey31: :+1: good point, I've made a small simplification in b910b3ba35ddea61361abba4088874414a5cf853. There's still a fair number of things we have to add to main so we'll have to keep this in mind while doing that. I'm thinking it's probably fine to leave the magic assertion since it's so short, what do you think though?

src/mem/buddy_allocator.rs
mtoohey31: Yeah this whole file could use some work, I'll try to do some cleanup.

src/main.rs
DerekCresswell: Sure, the magic is fairly self-contained and explanatory; it's one step. I think anything that's more than one step should  be wrapped up. Our target audience is young students!

src/mem/buddy_allocator.rs
mtoohey31: I've made an effort to clean things up in 5ce0a87f1404f7eff814fb93cd5a83152f40a7e1, let me know how that looks.

PR #4, opened 2024-02-02
Basic Threading Functionality

DerekCresswell:
This will stay fairly simple (useless) until interrupts, locks, and a few other things are created.

Implements the way to create and schedule threads within the OS.


DerekCresswell:
Tangling with the allocator was a night mare.

For the moment, we will need to finalize thread creation by allocating frames on the stack to be executed. I'll need to figure out how Rust expects a stack frame to be laid out. Part of this is creating the entry point for functions. PintOS puts the thread right into the 'switch_threads' function, but we may be able to put it just into a thunk (to call it's desired function).

The scheduler can probably exist as a global parameter (potentially with the ability to hot swap, like I believe the allocator can).

Once this system can actually run a single thread, we can make the kernel initialization enable the thread library and forgo control. We can put an idle / kernel thread into the queue to deal with non-urgent kernel tasks where there is low-volume of threads to run. The bulk of kernel work will just be handled in interrupts.

DerekCresswell:
This is ready for a (preliminary) review. The code current, when run, will show a context switch happening (and then halt).

I have tried to make the code fairly clean. But, of course, since this is low level code it may not be. 

There are many todo's left in the code and here is some of them:
* Set up the scheduling. The basics of it is ready, but locks and interrupts will be required to do it for real.
* Provide safety checks throughout. Some require locks or interrupts, but many just need to be implemented. Maybe also fixing up some of the panic / option spots.
* Tests. There are some obvious tests we can make. Such as the TCB stack functions and Scheduler functions.
* Stack types. We had mentioned using Box for the stack type but at first glance this required too many changes. We can maybe figure this out together / now that the code is more solid.
* Utilizing some more 'rust types' like.

One of the biggest things to discuss is what to do with the 'first' thread. In PintOS the kernel code that is initially running is converted into a thread. We have stated that we will not do that. It doesn't seems to work as is (the temporary code switches from 'not a thread' to a thread).

DerekCresswell:
Good to merge now (barring review).

mtoohey31 (approved):
Looks good! Like you've said, still some stuff to address in todo's but this is a good start! I'm going to merge this and rebase #7 on top of it.

PR #5, opened 2024-02-09
improve gdb debugging of kernel

mtoohey31:
This pull request makes it easier to debug the kernel with gdb. If you run `make run-qemu-gdb`, then run `rust-gdb` once qemu has started, you should be able to view source code, print locals, set breakpoints, etc.


serenayu11 (approved):
Works on my machine.

PR #7, opened 2024-03-02
Paging and Higher Half Kernel Address Space Layout via Trampoline

mtoohey31:
This pull request does two main things:
1. It implements some of the infrastructure for paging support.
2. It adds a "trampoline" (which is located in a separate section of the kernel's ELF file) that sets up bootstrap page tables before jumping into the kernel in such a way that all addresses (stack pointer, instruction pointer, the global allocator's pointers (both internal and those it returns upon allocation requests), etc.) are all virtual addresses above the 0xC0000000 offset. On startup, the kernel initializes the global allocator then replaces the bootstrap pages with ones allocated using the global allocator.

   Making the trampoline possible required some changes to the project structure. Since the trampoline code expects to be executed when located at physical addresses around 0x100000, whereas the kernel now expects to executed when located above 0xC0000000, we needed to put them in separate linking sections. We also needed separate copies of everything from the Rust `core` library for each due to the difference in expected address, and separate copies of some of our own code for functionality shared between the two (the video memory printer, paging, etc.)
   
   To support all of this, I broke the existing crate up into 3 separate parts:
   - `kidneyos-core` (located in the `core` directory) for stuff used by both the kernel and the trampoline.
   - `kidneyos-trampoline` (located in the `trampoline` directory) for the main trampoline code, and any helpers used exclusively by the trampoline.
   - `kidneyos` (located at the root of the repository) for the kernel itself.

   Using a separate crate for the trampoline is necessary because we need to build it into a static library so we can link two separate copies of shared functions into the resulting kernel. The separate core crate allows us to share code between the two without having to do complicated things to make the trampoline depend on the kidneyos crate's library output.

   While I was making these changes, I moved most of the Makefile's output to a `build` directory, since the top level directory was getting kind of cluttered.


mtoohey31:
The rebase promised here: https://github.com/KidneyOS/KidneyOS/pull/4#pullrequestreview-1918235519 has been completed, so this is ready for review!

src/paging.rs
```diff
@@ -0,0 +1,101 @@
...
+            write: true,
+        },
+    ];
+    assert!(regions
+        .iter()
+        .map(|Region { phys, .. }| phys.start)
+        .all(|i| i % PAGE_FRAME_SIZE == 0));
+
+    for Region { phys, virt, write } in regions {
+        for phys_addr in phys.clone().step_by(PAGE_FRAME_SIZE) {
```
DerekCresswell, line 55: I'm just commenting on style, but this chunk of code in particular seems very obtuse (potentially my small C brain's fault).

trampoline/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        mov cr3, {0}
+        mov {1}, cr0
+        or {1}, 0x80010000
+        mov cr0, {1}
+        ",
+        in(reg) page_directory as *mut PageDirectory as usize,
+        out(reg) _,
+    );
+
+    println!("Paging enabled!");
```
DerekCresswell, line 193: Why does both the trampoline and kernel have to 'enable' paging?

When booting, we do see these print statements getting all muddled together.

trampoline/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        call {}
+        ",
+        sym trampoline,
+        stack_size = const kidneyos_core::mem::OFFSET - kidneyos_core::mem::MAIN_STACK_SIZE,
+        options(noreturn),
+    )
+}
+
+#[allow(dead_code)]
+unsafe extern "C" fn trampoline(magic: usize, multiboot2_info: *mut Info) {
```
DerekCresswell, line 53: This function is very large and losing clarity because of that.

DerekCresswell (commented):
I understand the need for different crates, but I wonder if we really need the src and core/src separation. Looking at them, it makes the project confusing. At the very least we could name them differently. Kernel and something?

I don't mind how the main function is split up though. It's nice to sort of have memory 'off our minds' in that main function now. As well, having `kidneyos_core` rather than `crate` is nice.

We will need some documentation I think to delineate these sections and their purposes.

I'm really worried about the code duplication I see. Even things like those memory regions are just so similar, I would expect us to mess something up in the future.

DerekCresswell:
I think to dig into the code deeper, we could sit down together and go over things. I'd be happy to be more up to date on the nitty gritty of your work anyway.

DerekCresswell:
So, on boot we launch into 'trampoline land' where there is a dedicated page for all our info and addresses. We use this to set up the majority of our memory management / paging things.
Then, we can 'bounce' into the real kernel. All those trampoline pages are now part of virtual memory and can get mapped with a virtual address. So the kernel, and all it's allocators, then reference the new memory and virtual pages.

I personally will need some more convincing to make such large and seemingly drastic changes to the kernel. But I suspect you're right that it (or something similar) is needed.

trampoline/src/lib.rs
mtoohey31: The purpose of the trampoline is to set up the address space such that the kernel's virtual addresses can be above 0xC0000000. So it needs to enable paging so it can make those virtual addresses refer to the right physical addresses.

The kernel needs to do the same sort of thing again since we want to allocate the pages in the kernel's global allocator, which can't be initialized until the address space is already set up.

We can certainly swap out the term `enable` for something more appropriate. Maybe `activate` or something like that? When you say "getting all muddled together" do you mean that it's hard to tell what's what because they print the same things, as opposed to an actual bug in the video memory printer? The print statements are certainly not final, they're just there for now to provide info about what's happening. Would it help if there was a message marking the transition from trampoline to kernel?

mtoohey31:
> I think to dig into the code deeper, we could sit down together and go over things. I'd be happy to be more up to date on the nitty gritty of your work anyway.

I think that would help. Let me know when it would work for us to meet.

mtoohey31:
> I understand the need for different crates, but I wonder if we really need the src and core/src separation. Looking at them, it makes the project confusing. At the very least we could name them differently. Kernel and something?

This sort of multi-crate structure is relatively common in large Rust projects (for example, see [helix](https://github.com/helix-editor/helix) and [typst](https://github.com/typst/typst/tree/main/crates)).

It's possible we could try to cram everything into `src`, with part of the files in a library and part in a binary. But I would argue this would make things more confusing, not less, since you will then have to check the module hierarchy every time to determine whether a file is part of the library or the binary, instead of it being clear from the directory.

In terms of name, maybe `shared` instead of `core`? That's really it's purpose, to store anything shared between the trampoline and core.

Also worth noting, I don't expect core to ever grow beyond it's current size really. The trampoline currently accomplishes everything it needs to, so we shouldn't have to share any more code between the it and the kernel.

mtoohey31:
> I'm really worried about the code duplication I see. Even things like those memory regions are just so similar, I would expect us to mess something up in the future.

Is it mainly the code from `src/paging.rs` that's duplicated in `trampoline/src/lib.rs`? I do intend to get rid of that at some point soon. There is a significant difference between the two copies, in terms of how they get memory for the page tables. I'm thinking we can address this by making an abstraction on top of page directories/tables that's generic over the allocator, with the default being `Global`. Now that the pool allocator has been merged, we can use that in the trampoline. I was considering waiting until we're using the paging data structures more extensively in the kernel before writing that abstraction though, since we'll have a better idea of what it should look like then. Is that ok with you, or would you prefer that we do something to eliminate the code duplication before merging this?

Also, let me know if there's other things that are duplicated.

src/paging.rs
mtoohey31: Can you be more specific about what you find obtuse? I.e. is it the de-structuring of `Region`? Or the iteration over the range with `.step_by`?

trampoline/src/lib.rs
mtoohey31: Yeah it is pretty long lol, most of the length is the paging stuff though. Once we address the code duplication (futher discussion here: https://github.com/KidneyOS/KidneyOS/pull/7#issuecomment-1980117515) I'm thinking that should make it more manageable?

src/paging.rs
mtoohey31: Oh wait, reading this back now, did you mean the block below lines 54-55?

mtoohey31:
> There is a significant difference between the two copies, in terms of how they get memory for the page tables.

Just thought of one other important difference, the kernel's version has to translate the addresses which it uses to access page tables from virtual to physical. So we'll have to account for this when we build the abstraction discussed above, whenever it is that we decide to do that.

src/paging.rs
DerekCresswell: Oh yes, I meant this entire chunk! Maybe even just throw in some new lines ;)

trampoline/src/lib.rs
DerekCresswell: Muddled together: There is some slight graphical artifacts where this orignal 'Welcome to kidneyOS' gets overwritten. But I mainly mean that they say the same thing. Hard to tell what the logs are doing when they print duplicates.

On this, I think we should maybe make a quick 'kernel_log' function to use instead of println that formats the prints nicer (like when you see linux boot). That's purely aesthetic.

trampoline/src/lib.rs
DerekCresswell: Even just breaking it into functions where possible helps.

DerekCresswell:
LGTM

DerekCresswell (approved):

PR #8, opened 2024-03-03
fix memory leak in buddy allocator unit test

mtoohey31:
This pull request fixes a memory leak in the buddy allocator unit test. It also fixes the `test` makefile target so it tests both the binary and library tests for the `kidneyos` crate.

Memory leaks are still possible in this unit test when the test fails, but that's probably fine.

### DIFF
--- a/src/mem/buddy_allocator.rs
+++ b/src/mem/buddy_allocator.rs
@@ -222,7 +222,8 @@ mod tests {
 
     #[test]
     fn buddy_allocator_simple() -> Result<(), Box<dyn Error>> {
-        let region = Global.allocate(Layout::from_size_align(2 * KB, 2)?)?;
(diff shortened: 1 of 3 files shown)


src/mem/buddy_allocator.rs
```diff
@@ -259,6 +260,8 @@ mod tests {
 
         assert!(alloc.is_empty());
 
+        unsafe { Global.deallocate(region.as_non_null_ptr(), layout) };
```
Lei-Tin, line 263: O I didn't event notice that we didn't dealloc the region lol

But yes great fix

Lei-Tin (approved):
Cool!

Looking good

mtoohey31:
Great, thanks for the review!

PR #9, opened 2024-03-05
First implementation of simple spinlock

serenayu11:
This is the first implementation of a spinlock. 
Example of how to use the spinlock:
fn main() {
    
    // Initialize the SpinLock with some data.
    let counter = SpinLock::new(0);
    
    // Example of modifying the protected data.
    {
        let mut counter_guard = counter.lock();
        *counter_guard += 1;
        // Lock is automatically released when counter_guard goes out of scope.
    }

    // Example of reading the protected data.
    {
        let counter_guard = counter.lock();
        println!("Counter value is: {}", *counter_guard);
        // Lock is automatically released when counter_guard goes out of scope.
    }

    // Further processing...
}

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -9,7 +9,7 @@
 mod interrupt_descriptor_table;
 mod mem;
 mod paging;
-mod synch;
+mod sync;
 #[allow(unused)]
 mod threading;
 
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -9,6 +9,7 @@
 mod interrupt_descriptor_table;
 mod mem;
 mod paging;
+mod synch;
 #[allow(unused)]
 mod threading;
 
--- a/kernel/src/sync.rs
+++ b/kernel/src/sync.rs
@@ -13,6 +13,8 @@ unsafe impl<T> Sync for SpinLock<T> {}
 unsafe impl<T> Send for SpinLock<T> {}
 
 impl<T> SpinLock<T> {
+    #![allow(unused)]
+
     // Creates a new spinlock.
     pub const fn new(data: T) -> SpinLock<T> {
         SpinLock {
@@ -73,6 +75,8 @@ unsafe impl<T> Sync for InterruptLock<T> {}
 unsafe impl<T> Send for InterruptLock<T> {}
 
 impl<T> InterruptLock<T> {
+    #![allow(unused)]
+
     // Creates a new InterruptLock.
     pub const fn new(data: T) -> InterruptLock<T> {
         InterruptLock {
--- a/kernel/src/sync.rs
+++ b/kernel/src/sync.rs
@@ -1,7 +1,6 @@
-use core::sync::atomic::{AtomicBool, Ordering};
 use core::cell::UnsafeCell;
 use core::ops::{Deref, DerefMut};
-
+use core::sync::atomic::{AtomicBool, Ordering};
 
 // A simple spinlock.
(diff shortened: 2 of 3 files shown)


src/synch.rs
```diff
@@ -0,0 +1,68 @@
+#![no_std]
+#![feature(core_intrinsics)]
+
```
mtoohey31, line 3:
```suggestion
```

Looks like this can safely be removed?

src/synch.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct SpinLock<T> {
+    lock: AtomicBool,
+    data: UnsafeCell<T>,
+}
+
+#[derive(Debug)]
+pub enum MutexError {
+    Poisoned,
+}
+
```
mtoohey31, line 18:
```suggestion
```

This also appears unused.

src/synch.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub const fn new(data: T) -> SpinLock<T> {
+        SpinLock {
+            lock: AtomicBool::new(false),
+            data: UnsafeCell::new(data),
+        }
+    }
+
+    // Acquires the spinlock, spinning until the lock is obtained.
+    pub fn lock(&self) -> SpinLockGuard<T> {
+        while self.lock.compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed).is_err() {
```
mtoohey31, line 34:
```suggestion
        while self
            .lock
            .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
            .is_err()
        {
```

Sorry, I haven't gotten around to adding a rustfmt check as part of a GitHub action, but if you make this change then the file should match what rustfmt expects.

mtoohey31 (changes requested):
Looks great!

Just a few small suggestions. Also, I think we need to add a mod declaration to `lib.rs` before we can use this in other files:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 7a42933..0c3ce64 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,7 @@ pub mod constants;
 pub mod macros;
 pub mod mem;
 pub mod serial;
+pub mod synch;
 pub mod video_memory;
 
 extern crate alloc;
```

src/synch.rs
serenayu11: I think we might need it in the future? So it can be used to indicate that a panic occurred while the lock was held.

mtoohey31: I think mutex poisoning is mainly useful when you’re using multiple threads/cores, and one panics, but the others are still around to clean things up. My thought process is that since we’re using only a single core, if a panic occurs then there’s no chance of recovery. Maybe it would be possible to define some error recovery logic in the panic handler, but I think it would be extremely difficult to figure out how to restore the OS to a consistent state after a panic, so imo the best approach would be to do our best not to panic, and not worry about attempting to recover from panics. Does that seem reasonable?

mtoohey31:
Thanks for making those changes! One other small nitpick I forgot to mention in the initial review, could we rename from synch to sync to be consistent with the naming of sync in core?

mtoohey31 (approved):
Looks great! I've just formatted things, and I think they're ready to be merged. Now I just have to figure out how to deal with the conflicts... will try to get that done soon!

mtoohey31:
Ok, I think I managed to rebase everything without making too much of a mess, so I'm merging things.

PR #10, opened 2024-03-05
Add github action

mtoohey31:
This pull request adds a GitHub action to the repository that will allow us to check that clippy lints are addressed, code is formatted, tests are passing, and that the ISO builds.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -2,5 +2,5 @@
 
 set -euo pipefail
 
-cargo clippy --quiet
+cargo clippy --quiet -- -D warnings
 find src -name '*.rs' -print0 | xargs -0 cargo fmt -- --quiet --check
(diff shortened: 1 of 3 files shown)


Lei-Tin (approved):
Looks great to me! 
Seems like they are successfully calling the tests and are building normally

However, I'm not quite sure what the formatting task (rustfmt) will show when the formatting is not adhereing to standards. Currently, I think it just displays nothing. (I think because all the files are currently formatted well?) 

(Side note: I actually didn't know GitHub Actions can display colors so I was quite surprised seeing you throw the `CARGO_TERM_COLOR` everywhere lol)

mtoohey31:
Thanks!

> However, I'm not quite sure what the formatting task (rustfmt) will show when the formatting is not adhereing to standards. Currently, I think it just displays nothing. (I think because all the files are currently formatted well?)

Yeah it displays nothing right now because there’s no changes necessary. When it fails, it displays a diff of what needs to be changed. There seems to be a bug where it displays the diff twice, though I haven’t looked into it because it’s easier to fix things by just running `cargo fmt` instead of applying the diff manually.

PR #14, opened 2024-03-07
Scheduling

DerekCresswell:


DerekCresswell:
As discussed in the meeting, we need to observe ownership rules when placing TCBs into the scheduler / running_thread. We could nicely do this in the end of `switch_threads` after performing the context switch. However, the first time a thread is run, it leaves the context switch and enters `prepare_thread` rather than finishing `switch_threads`. We will need to either make the thread start running midway through a function (sounds awful) or just manage to deal with the scheduler sooner and break some rules by using the pointer values after giving away ownership of what they came from.

DerekCresswell:
This is being far more of a headache then I want it to be. But great progress.

The only thing left for me to do I place my TCB pointers correctly on the stack for run_thread. I forgot orignally that run_thread is not naked and so messes with the stack more than I wanted it to. Should be one more little session to fix,

DerekCresswell:
I do need to figure out how to place our `*mut ThreadControlBlock` type into the scheduler, Rust is complaining about that...

DerekCresswell:
@mtoohey31 Maybe I can get a set of eyes on this because it's now proving to be far more headache than it should.

This works to switch to a thread, back, and then almost back again. I believe it has to do with my use of Box::into_raw and Box::from_raw. I feel pretty confident that the parts to actually switch around threads is correct as I spent a lot of time painstakingly stepping through GDB.
However, the boxing seems to be messing with addresses some. I feel you might have a good suggestion.

DerekCresswell:
Per advice from matthew:
* Just box up all TCBs to begin with. Should hopefully make less headache in the future.
* EBP is part of the callee side of calling conventions, our context switch might need to deal with it a little more clearly. It might be messing up after we return from the context switch from a previously running thread.

DerekCresswell:
We are also going to mark `context_switch`, `prepare_thread` and `run_thread` as `extern "C"` to hopefully make the calling conventions a bit more normalized.

DerekCresswell:
This boxing of everything seems good. Next I will make all naked functions extern C as well. Currently it seems like we are not grabbing the correct instruction pointer all the time. Could be remedied by this change, but otherwise I will need to dig deeper.

DerekCresswell:
Ahh finally! She works well and good.

Going to fix up merging stuff but this should finally work rather robustly. There is some documentation included, but I am happy to defend any choices made too.

In review, pay attention to things that are not very `rust like` as I'm still learning some of those things.
For instance, I'd love a better suggestion for how to wrap accessing the scheduler or changing the status of the running thread more conveniently (must be done unsafe with no interrupts and in quite a few spots).

DerekCresswell:
The error being shown in the logs does not appear in my console. Seems to be referencing an old file that is now gone. Not sure what to do about that. I got a similar one when merging.

DerekCresswell:
Matthew, if you would give this a review that'd be lovely.

Don't merge it quite yet. Once you review I will move interrupts / locking  into the equation as a last step.

kernel/src/threading/mod.rs
```diff
@@ -4,8 +4,29 @@ pub mod thread_control_block;
 pub mod thread_functions;
```
mtoohey31, line 4: If you want to be careful about the API you're exposing, one good step would be to remove the `pub`s here (and maybe go through the rest of this file and check if you can remove any other `pub` specifiers). In the sub-`mod`s pub is fine once you remove stuff here since it will only be made visible to stuff in the parent. If there are specific things in the sub-`mod`s that you want to make visible, you can do:

```rust
pub use context_switch::foo;
```

kernel/src/threading/mod.rs
```diff
@@ -4,8 +4,29 @@ pub mod thread_control_block;
...
 
 use crate::println;
+use alloc::boxed::Box;
+
+use crate::threading::scheduling::initialize_scheduler;
 use crate::threading::thread_control_block::{ThreadControlBlock, Tid};
 
+use self::scheduling::{scheduler_yield, SCHEDULER};
```
mtoohey31, line 12:
```suggestion
use thread_control_block::{ThreadControlBlock, Tid};
use scheduling::{initialize_scheduler, scheduler_yield, SCHEDULER};
```

kernel/src/threading/thread_functions.rs
```diff
@@ -8,56 +14,55 @@ pub type ThreadFunction = fn() -> ();
...
 
 #[repr(C, packed)]
 pub struct PrepareThreadContext {
-    eip: usize, // Should always be set to &run_thread.
+    entry_function: *const ThreadFunction,
 }
 
 impl PrepareThreadContext {
-    pub fn create() -> Self {
+    pub fn create(entry_function: ThreadFunction) -> Self {
```
mtoohey31, line 63: Can we rename this to `new`? That's the typical rust convention, unless there's a reason why this should be distinguished from what typical `new` functions do?

kernel/src/threading/mod.rs
```diff
@@ -32,10 +55,47 @@ pub fn thread_system_initialization() {
  * Enables preemptive scheduling.
  * Thread system must have been previously enabled.
  */
-pub fn thread_system_start() {
+pub fn thread_system_start() -> ! {
+    // SAFETY: Interrupts must be disabled.
```
mtoohey31, line 59: Might want to leave a TODO to just `assert!` that interrupts are disabled so we're sure in the future, once there's a function for checking the interrupt state.

kernel/src/threading/scheduling/fifo_scheduler.rs
```diff
@@ -0,0 +1,33 @@
+use crate::alloc::boxed::Box;
+use alloc::collections::VecDeque;
```
mtoohey31, line 2:
```suggestion
use alloc::{boxed::Box, collections::VecDeque};
```

kernel/src/threading/scheduling/fifo_scheduler.rs
```diff
@@ -0,0 +1,33 @@
...
+use super::super::{ThreadControlBlock, Tid};
+
+use super::scheduler::Scheduler;
+
+pub struct FIFOScheduler {
+    ready_queue: VecDeque<Box<ThreadControlBlock>>,
+}
+
+// SAFETY: Schedulers should be run with interrupts disabled.
+unsafe impl Sync for FIFOScheduler {}
```
mtoohey31, line 13:
```suggestion
```

This should be removed. It is precisely because `FIFOScheduler` should be run with interrupts disabled that it must not implement `Sync`. Anywhere you need to stick a `FIFOScheduler` that it needs to be sync should be replaced with `InterruptLock<FIFOScheduler>`.

kernel/src/threading/scheduling/mod.rs
```diff
@@ -0,0 +1,38 @@
+pub mod fifo_scheduler;
+pub mod scheduler;
```
mtoohey31, line 2: Same advice here about removing `pub` and intentionally `pub use`'ing things that need to be accessible elsewhere.

```suggestion
mod fifo_scheduler;
mod scheduler;
```

kernel/src/threading/scheduling/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use self::fifo_scheduler::FIFOScheduler;
+use self::scheduler::Scheduler;
```
mtoohey31, line 5:
```suggestion
use fifo_scheduler::FIFOScheduler;
use scheduler::Scheduler;
```

kernel/src/threading/scheduling/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+use alloc::boxed::Box;
+
+use super::context_switch::switch_threads;
+use super::thread_control_block::ThreadStatus;
+use super::RUNNING_THREAD;
+
+pub static mut SCHEDULER: Option<Box<dyn Scheduler>> = None;
+
+pub fn initialize_scheduler() {
+    // SAFETY: Interrupts should be off.
```
mtoohey31, line 16: Maybe add an assert or a TODO to add an assert here too. There are probably other places so maybe just grep the code for similar comments. (I might also be guilty of not asserting where I should've in some of the existing code, feel free to bug me to do that if you're finding lots that are my fault.)

kernel/src/threading/thread_control_block.rs
```diff
@@ -24,23 +25,18 @@ pub enum ThreadStatus {
     Dying,
 }
 
+#[repr(C, packed)]
 pub struct ThreadControlBlock {
+    stack_pointer: NonNull<u8>, // Must be kept as the top of the struct so it has the same address as the TCB.
+    stack_pointer_bottom: NonNull<u8>, // Kept to avoid dropping the stack and to detect overflows.
+    // TODO: Can we just leak this and not hold it? Keep size instead, then kernel thread won't have a dangling pointer.
```
mtoohey31, line 32: Can you check again if this still breaks things if you don't keep it? [NonNull](https://doc.rust-lang.org/src/core/ptr/non_null.rs.html#74-76) doesn't have any drop implementation and it just holds a raw pointer, so dropping it shouldn't be deallocating anything.

kernel/src/threading/context_switch.rs
```diff
@@ -1,50 +1,71 @@
 use crate::threading::ThreadControlBlock;
 use core::mem::align_of;
 
+use alloc::boxed::Box;
+
+use super::{scheduling::SCHEDULER, thread_control_block::ThreadStatus, RUNNING_THREAD};
+
 /**
```
mtoohey31, line 8: A style note, it looks like `/**` does technically count as a rust doc comment, but basically nobody uses the block comment style. I'm not going to be too picky about this so you can leave these if you'd like, but just a heads up in case you end up working on any other rust codebases in the future, normally people write doc comments with `///`.

kernel/src/threading/thread_functions.rs
```diff
@@ -1,3 +1,9 @@
+use super::scheduling::SCHEDULER;
+use super::thread_control_block::{ThreadControlBlock, ThreadStatus};
+use super::RUNNING_THREAD;
+
+use crate::alloc::boxed::Box;
```
mtoohey31, line 5:
```suggestion
use alloc::boxed::Box;
```

kernel/src/threading/thread_functions.rs
```diff
@@ -8,56 +14,55 @@ pub type ThreadFunction = fn() -> ();
...
-    }
+    panic!("Thread exited incorrectly.");
 }
 
 /**
  * A wrapper function to execute a thread's true function.
  */
-fn run_thread(function: ThreadFunction) {
-    // TODO: Safety checks.
+#[no_mangle]
```
mtoohey31, line 28: What is the `no_mangle` here for?

kernel/src/threading/thread_functions.rs
```diff
@@ -8,56 +14,55 @@ pub type ThreadFunction = fn() -> ();
...
-    eip: usize, // Should always be set to &run_thread.
+    entry_function: *const ThreadFunction,
 }
 
 impl PrepareThreadContext {
-    pub fn create() -> Self {
+    pub fn create(entry_function: ThreadFunction) -> Self {
         Self {
-            eip: run_thread as usize,
+            entry_function: entry_function as *const ThreadFunction,
```
mtoohey31, line 65: Not 100% sure but I think this might be wrong. By casting `entry_function` to a `*const ThreadFunction`, I believe you're saying it's a pointer to a function pointer, which it isn't: it's the function pointer itself. I think you can just drop the cast and change the type of the field to `ThreadFunction`.

kernel/src/threading/thread_functions.rs
```diff
@@ -66,12 +71,21 @@ impl PrepareThreadContext {
  * This function is used to clean up a thread's arguments and call into `run_thread`.
  */
 #[naked]
+#[no_mangle]
```
mtoohey31, line 74: Also wondering what the `no_mangle` is about here.

kernel/src/threading/thread_functions.rs
```diff
@@ -81,11 +95,11 @@ unsafe extern "C" fn prepare_thread() {
...
-    edi: usize, // Destination index.
-    esi: usize, // Source index.
-    ebx: usize, // Base (for memory access).
-    ebp: usize, // Stack base pointer.
-    eip: usize, // Index pointer.
+    edi: usize,                 // Destination index.
+    esi: usize,                 // Source index.
+    ebx: usize,                 // Base (for memory access).
+    ebp: usize,                 // Stack base pointer.
+    eip: *const ThreadFunction, // Instruction pointer (determines where to jump after the context switch).
```
mtoohey31, line 102: I think this should also probably be just plain `ThreadFunction`.

kernel/src/threading/thread_functions.rs
```diff
@@ -105,7 +119,7 @@ impl SwitchThreadsContext {
             esi: 0,
             ebx: 0,
             ebp: 0,
-            eip: prepare_thread as usize,
+            eip: prepare_thread as *const ThreadFunction,
```
mtoohey31, line 122: Same as above.

kernel/src/main.rs
```diff
@@ -15,7 +15,7 @@ mod threading;
 
 extern crate alloc;
 
-use crate::threading::thread_system_initialization;
+use crate::threading::{thread_system_initialization, thread_system_start};
```
mtoohey31, line 18:
```suggestion
use threading::{thread_system_initialization, thread_system_start};
```

mtoohey31 (commented):
Super excited that this is working! Mostly style comments, as requested, plus some questions. Thanks!

kernel/src/threading/scheduling/fifo_scheduler.rs
DerekCresswell: Remnant from a compiler warning of old.

kernel/src/threading/thread_functions.rs
DerekCresswell: No mangle makes it so the rust compiler must keep the same names for functions. Though, I saw some sources suggest that us using the call and sym may mean that rust can mangle names, but still get everything lined up as needed.

kernel/src/threading/scheduling/fifo_scheduler.rs
DerekCresswell: MM, it does complain  without, but that seems to stem from the NonNull pointer for the stack. I will address this with that other concern.

kernel/src/threading/thread_functions.rs
mtoohey31: Yeah `sym` seems to do the trick there. I think we have some other code that uses `sym` without `no_mangle`.

kernel/src/threading/thread_functions.rs
mtoohey31: Yeah I believe so. Like you can cast directly from a function identifier to a `usize`.

kernel/src/threading/mod.rs
DerekCresswell: I believe that everything is marked as private as it can be for now.

Certain functions, like `switch_threads` are public but things like `context_switch` are kept private.

DerekCresswell:
I've been able to leave all of the thread_functions as simply thread functions and it all works well. Though, the entry functions must be marked as extern C. I don't think this is an issue as we expect outside C programs to be run as user programs.

DerekCresswell:
I am going to make separate PR's for the Stack type and handlers in the IDT (unless Matthew beats me to that one). I want this to get in main sooner rather than later.

mtoohey31 (dismissed):
> I am going to make separate PR's for the Stack type and handlers in the IDT (unless Matthew beats me to that one). I want this to get in main sooner rather than later.

Will the separate PR for the stack type address these: https://github.com/KidneyOS/KidneyOS/pull/14#discussion_r1529449677 and https://github.com/KidneyOS/KidneyOS/pull/14#discussion_r1529456514? If so, would you mind replacing the safety comment on the first with a TODO saying that we plan to remove that? Other than that, this looks good to me.

DerekCresswell:
Yes, I will address those in the stack type, it's a little more involved but it's all internal so I'm not worried about this being in main.

I believe that fault is actually the locking issue #18 I fixed (this is how I found it). I will re-merge main into this and double check.

DerekCresswell:
Yeah, it was the just the lock (and the fact that I had my checks backwards haha). Ready minus my comment about the `use crate::`.

DerekCresswell:
Hmm, I'm trying to rebase this to remove some of the WIP commits, but it is soo confusing and weird, I think due to the number of times I merged main... : (

DerekCresswell:
Ok! I will stop touching this now

kernel/src/threading/context_switch.rs
```diff
@@ -1,32 +1,39 @@
...
         panic!("Cannot switch to a non-ready thread.");
     }
 
-    // Ensure that the previous thread is no longer running.
-    if (*switch_from).status == ThreadStatus::Running {
-        panic!("The thread to switch out of must no longer be in the running state.");
+    // Ensure that the previous thread is running.
+    if (*switch_from).status != ThreadStatus::Running {
+        panic!("The thread to switch out of must be in the running state.");
     }
```
mtoohey31, line 32: Sorry should've caught this in the earlier review but you can use `assert!` here instead of writing the if statement. `assert!` optionally allows a message as the second argument, and panics with that message if the assertion fails.

mtoohey31 (approved):
Looks good! Just one small tip that could make things shorter, but it won't change the behaviour so it's your call as to whether to make that change. Feel free to merge when it's ready. Thanks!

PR #15, opened 2024-03-09
Add PageManager

mtoohey31:
`PageManager` provides a more convenient abstraction over page directories and tables. I'll update this description with something more detailed once the changes are finished.

### DIFF
--- a/kernel/src/interrupt_descriptor_table.rs
+++ b/kernel/src/interrupt_descriptor_table.rs
@@ -62,7 +62,6 @@ unsafe extern "C" fn page_fault_handler() -> ! {
     );
 }
 
-#[allow(clippy::cast_possible_truncation)]
 static mut IDT_DESCRIPTOR: IDTDescriptor = IDTDescriptor {
     size: size_of::<[GateDescriptor; IDT_LEN]>() as u16 - 1,
     offset: 0, // Will fetch pointer and set at runtime below.
@@ -73,20 +72,17 @@ static mut IDT_DESCRIPTOR: IDTDescriptor = IDTDescriptor {
 /// Can only be executed within code that expects the interrupt handlers to be
 /// defined as they are above.
 pub unsafe fn load() {
-    // u32 is wide enough since we're writing 32-bit code.
-    #![allow(clippy::fn_to_numeric_cast_with_truncation)]
-
     IDT_DESCRIPTOR.offset = IDT.as_ptr() as u32;
 
     for gate_descriptor in &mut IDT {
         *gate_descriptor = GateDescriptor::default()
-            .with_offset(unhandled_handler as u32)
+            .with_offset(unhandled_handler as usize as u32)
             .with_segment_selector(0x8)
             .with_gate_type(u4::new(0xE))
             .with_descriptor_privilege_level(u2::new(3))
             .with_present(true);
     }
-    IDT[0xE] = IDT[0xE].with_offset(page_fault_handler as u32);
+    IDT[0xE] = IDT[0xE].with_offset(page_fault_handler as usize as u32);
 
     asm!("lidt [{}]", sym IDT_DESCRIPTOR);
 }
--- a/shared/src/paging.rs
+++ b/shared/src/paging.rs
@@ -158,10 +158,11 @@ impl<A: Allocator> PageManager<A> {
             virt_start,
             len,
             write,
+            user,
         } in mapping_ranges
         {
             // SAFETY: res has not yet been loaded.
-            unsafe { res.map_range(phys_start, virt_start, len, write) };
+            unsafe { res.map_range(phys_start, virt_start, len, write, user) };
         }
 
         res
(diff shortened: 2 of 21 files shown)


mtoohey31:
Ok I've got PSE working so this should be ready for review.

There's one thing that I know is a little ugly, which is the use of arch cfg's in the definition of PSE_SUPPORTED. I'm going to see if I can get things configured so all our checks and language servers can run against our exact target, but until then the workaround there is the best we can do.

Also, if using lazy static seems a bit overcomplicated, let me know.

DerekCresswell:
It's a sizable chunk of code, but it all looks very well done. I swear you're just inventing new rust syntax though!

Not sure why my code seems to not be in an x86 environment but that limits my abilities to test the changes.

shared/src/paging.rs
```diff
@@ -109,15 +110,466 @@ pub struct PageTableEntry {
...
+        or {0}, {mask}
+        mov cr0, {0}
+        ",
+        out(reg) _,
+        mask = const MASK,
+        options(nomem, nostack),
+    );
+}
+
+lazy_static! {
```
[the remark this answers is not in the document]
mtoohey31: 🤦 probably lol, that would make it a lot simpler, and I can't really think of any downsides? Totally didn't even think about it!

mtoohey31:
> It's a sizable chunk of code, but it all looks very well done. I swear you're just inventing new rust syntax though!
> 
> 
> 
> Not sure why my code seems to not be in an x86 environment but that limits my abilities to test the changes.

Great! Thanks for the review!

Sounds like I might've changed some stuff that broke the build on other peoples' systems by using the `arch::86` crate? Not really sure why it would be panicking though... If it's ok with you, the fastest way to figure this out might just be for me to take a look at how the build runs on your laptop at the meeting tomorrow.

Also related to the above, just wondering: do you also get errors showing up in VScode after these changes? I'm mainly wondering about some of the changes in the `paging.rs` file. I changed the target platform for my language server (those are the changes in `.helix/languages.toml`) and I'm wondering if we might want to do the same thing for VScode.

shared/src/paging.rs
```diff
@@ -109,15 +110,466 @@ pub struct PageTableEntry {
...
+            pse: bool,
+        }
+
+        #[cfg(arch = "x86")]
+        {
+            let core::arch::x86::CpuidResult { edx, .. } = unsafe { core::arch::x86::__cpuid(0) };
+            CPUIDEdx::new_with_raw_value(edx).pse()
+        }
+        #[cfg(not(arch = "x86"))]
+        unimplemented!()
```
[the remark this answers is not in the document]
mtoohey31: This is resolved in dd25ae07b1214d18e5f23b0a7dc4b3e3de6cbdd4.

kernel/src/interrupt_descriptor_table.rs
```diff
@@ -73,20 +72,17 @@ static mut IDT_DESCRIPTOR: IDTDescriptor = IDTDescriptor {
...
         *gate_descriptor = GateDescriptor::default()
-            .with_offset(unhandled_handler as u32)
+            .with_offset(unhandled_handler as usize as u32)
             .with_segment_selector(0x8)
             .with_gate_type(u4::new(0xE))
             .with_descriptor_privilege_level(u2::new(3))
             .with_present(true);
     }
-    IDT[0xE] = IDT[0xE].with_offset(page_fault_handler as u32);
+    IDT[0xE] = IDT[0xE].with_offset(page_fault_handler as usize as u32);
```
DerekCresswell, line 85: Aren't these redundant casts?

shared/src/paging.rs
```diff
@@ -109,15 +110,460 @@ pub struct PageTableEntry {
...
+    /// any existing pointers to refer to anything they shouldn't.
+    pub unsafe fn load(&self) {
+        let root_phys_addr = self.root.as_ptr() as usize - self.phys_to_alloc_addr_offset;
+        unsafe { asm!("mov cr3, {}", in(reg) root_phys_addr, options(nomem, nostack)) };
+    }
+
+    /// Returns whether these page tables are loaded.
+    pub fn is_loaded(&self) -> bool {
+        let current_root: usize;
+        unsafe { asm!("mov {}, cr3", out(reg) current_root, options(nomem, nostack)) };
```
DerekCresswell, line 190: Not sure if registers count as mem to `nomem`.

shared/src/paging.rs
```diff
@@ -109,15 +110,460 @@ pub struct PageTableEntry {
...
+        // Check if PSE is already enabled (from the trapoline, if we're running
+        // in the kernel).
+        #[bitfield(u32, default = 0)]
+        struct CR4 {
+            #[bit(4, rw)]
+            pse: bool,
+        }
+
+        let cr4: u32;
+        unsafe { asm!("mov {}, cr4", out(reg) cr4, options(nomem, nostack)) };
```
DerekCresswell, line 505: Same...

DerekCresswell (approved):

kernel/src/interrupt_descriptor_table.rs
mtoohey31: This is to avoid triggering the `clippy::fn_to_numeric_cast` lint. I could disable the lint entirely but then it wouldn't warn about casting functions to other numeric types that would actually truncate.

shared/src/paging.rs
mtoohey31: It doesn't look like they do: https://doc.rust-lang.org/reference/inline-assembly.html#options.

shared/src/paging.rs
mtoohey31: See above.

shared/src/paging.rs
mtoohey31: Oh actually, while re-reading the spec I realized I probably shouldn't include `nomem` on stuff that changes the state of paging stuff. I've fixed that in f5ce4ea8d1ec40d407e48ddec3b2ed397caa87ab.

PR #16, opened 2024-03-20
Nested Lock Implementations

DerekCresswell:


DerekCresswell:
Unused warnings can be ignored. These locks will be used elsewhere.

serenayu11 (approved):
Looks good

mtoohey31:
I've pushed a small change so that clippy won't break CI. If that change looks ok then feel free to merge the branch!

PR #17, opened 2024-03-20
Fix Makefile Bug

ff-zhang:
Fixes an issue when building on MacOS ARM chips (due to `$(OBJCOPY)$`) by replacing the instances of it with `i686-unknown-linux-gnu-objcopy`.


mtoohey31 (approved):
Thanks, looks good!

PR #18, opened 2024-03-23
Hotfix: Properly restore stack pointer in ASM

DerekCresswell:
Utilizes the `pushfd` command over the `pushf` command to ensure that our stack pointer is moved by 4 bytes at a time. Uses the `popfd` command to properly restore the stack pointer.


mtoohey31 (approved):
Good catch, thanks!

PR #19, opened 2024-03-24
VMA and ELF file parsing

serenayu11:
First simple implementation of VMA, currently it only works for assigning permission to a region.
First implementation of ELF file parsing, within the function, there's a loop that iterates over the program headers. For each header, if it's a loadable segment (PT_LOAD), the code would load the segment into memory. The core::intrinsics::breakpoint() call is a placeholder for where we would implement the actual memory-loading logic. Process.rs is more like pseudocode and we need to revise and replace placeholder functions with our own threading function calls, we can delete the file for this pull request.


kernel/src/user_program/elf_loader.rs
```diff
@@ -0,0 +1,133 @@
+extern crate core;
+
+use core::slice;
+use virtual_memory_area::VmAreaStruct;
+use virtual_memory_area::VmFlags;
```
mtoohey31, line 5:
```suggestion
use super::virtual_memory_area::{VmAreaStruct, VmFlags};
```

The first two statements are unnecessary, and the second two need the `super::` prefix and can be combined.

kernel/src/user_program/elf_loader.rs
```diff
@@ -0,0 +1,133 @@
...
+    PtStack = 0x6474e551, // Stack segment.
+}
+
+const ELF_MAGIC_NUMBER: [u8; 4] = [0x7F, b'E', b'L', b'F'];
+
+// Dummy function to simulate reading an ELF file from somewhere into a byte slice.
+// TODO: replace this with actual file reading logic once we have file system implemented.
+fn read_elf_file() -> &'static [u8] {
+    &include_bytes!("path/to/your/elf/file")[..]
+}
```
mtoohey31, line 58: This won't compile since the path doesn't exist. Could we remove this for now? We also don't really need a function to wrap this, we can just use `include_bytes!` inline.

kernel/src/user_program/elf_loader.rs
```diff
@@ -0,0 +1,133 @@
...
+
+// Function to verify ELF header
+fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
```
mtoohey31, line 61: `ElfError` doesn't appear to be defined anywhere... can you add a definition?

kernel/src/user_program/virtual_memory_area.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub read: bool,
+    pub write: bool,
+    pub execute: bool,
+    pub shared: bool,
+    pub private: bool,
+    // TODO: add other necessary flags.
+}
+
+impl VmAreaStruct {
+    fn new(vm_start: usize, vm_end: usize, flags: VmFlags) -> Self {
```
mtoohey31, line 70: This needs to be marked as `pub` for the usage in `elf_loader.rs` to work.

kernel/src/user_program/elf_loader.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    // Iterate over program headers
+    let ph_offset = header.e_phoff as usize;
+    let ph_size = header.e_phentsize as usize;
+    for i in 0..header.e_phnum as usize {
+        let ph = unsafe {
+            &*(elf_data.as_ptr().offset((ph_offset + i * ph_size) as isize) as *const Elf32Phdr)
+        };
+
+        if ph.p_type == SegmentType::PtLoad {
```
mtoohey31, line 113:
```suggestion
        if ph.p_type == SegmentType::PtLoad as u32 {
```

You need to cast this before performing the comparison or else the two operands have different types.

mtoohey31 (changes requested):
It looks like the new files weren't part of the module tree because of a missing `mod` declaration. I added the declaration in fa48a6275233e5cb77b8e28ad7a8f3f5eb53bb91 but now there are a bunch of build errors (see the output of https://github.com/KidneyOS/KidneyOS/actions/runs/8414174890/job/23037462788). I added a few comments suggesting how to fix some of the issues, but I haven't gone through everything.

kernel/src/user_program/elf_loader.rs
serenayu11: Sure sure, I can make the change

kernel/src/user_program/elf_loader.rs
serenayu11: Yes, just did it

serenayu11:
All checks have passed, please help to check if we can merge pull request, thanks :)

kernel/src/user_program/elf_loader.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Main function to load the ELF binary
+#[allow(unused)]
+fn load_elf(elf_data: &[u8]) {
+    let header = unsafe { &*(elf_data.as_ptr() as *const Elf32Ehdr) };
+    let mut vm_areas = Vec::new();
+
+    // Verify ELF header
+    verify_elf_header(header);
```
mtoohey31, line 112: If this function errors, it won't have any effect since we're not checking the return value. We should change the return type of `load_elf` to `Result<(), ElfError>` and check the result by adding `?` after this function call. You'll need to add `Ok(())` to the end of the function too.

mtoohey31 (changes requested):
Just one last suggestion.

mtoohey31 (approved):
Thanks for making those changes, approved!

serenayu11:
> Thanks for making those changes! I've approved the PR.

Thanks!

PR #20, opened 2024-03-25
Add segment validation for each program header.

serenayu11:
The function validate_segment() is added, it is responsible for verifying that a segment described by a program header in an ELF file is valid and can be safely loaded into memory for execution.

### DIFF
--- a/kernel/src/user_program/elf_loader.rs
+++ b/kernel/src/user_program/elf_loader.rs
@@ -1,5 +1,6 @@
 use super::virtual_memory_area::{VmAreaStruct, VmFlags};
 use alloc::vec::Vec;
+use kidneyos_shared::mem::{OFFSET, PAGE_FRAME_SIZE};
 
 #[repr(C)]
 #[derive(Debug)]
@@ -64,6 +65,8 @@ const PF_W: u32 = 2; // Writable.
 const PF_R: u32 = 4; // Readable.
 
 const ELF_MAGIC_NUMBER: [u8; 4] = [0x7F, b'E', b'L', b'F'];
+const PHYS_BASE: *const () = OFFSET as *const ();
+const PGMASK: usize = OFFSET - 1;
 
 // Function to verify ELF header
 fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
@@ -102,6 +105,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
     Ok(())
 }
 
+#[allow(unused)]
+fn is_kernel_vaddr(vaddr: *const ()) -> bool {
+    unsafe { vaddr >= PHYS_BASE }
+}
+
+#[allow(unused)]
+fn is_user_vaddr(vaddr: *const ()) -> bool {
+    unsafe { vaddr < PHYS_BASE }
+}
+
+fn validate_segment(phdr: &Elf32Phdr, file_data: &[u8]) -> bool {
+    // p_offset and p_vaddr must have the same page offset.
+    if (phdr.p_offset as usize & PGMASK) != (phdr.p_vaddr as usize & PGMASK) {
+        return false;
+    }
+
+    // p_offset must point within FILE.
+    if phdr.p_offset as usize > file_data.len() {
+        return false;
+    }
+
+    // p_memsz must be at least as big as p_filesz.
+    if phdr.p_memsz < phdr.p_filesz {
+        return false;
+    }
+
+    // The segment must not be empty.
+    if phdr.p_memsz == 0 {
+        return false;
+    }
+
+    // The virtual memory region must both start and end within the
+    // user address space range.
+    if !is_user_vaddr(phdr.p_vaddr as *const ()) {
+        return false;
+    }
+    if !is_user_vaddr((phdr.p_vaddr as usize + phdr.p_memsz as usize) as *const ()) {
+        return false;
+    }
+
+    // The region cannot "wrap around" across the kernel virtual
+    // address space.
+    if (phdr.p_vaddr as usize + phdr.p_memsz as usize) < (phdr.p_vaddr as usize) {
+        return false;
+    }
+
+    // Disallow mapping page 0.
+    if (phdr.p_vaddr as usize) < (PAGE_FRAME_SIZE as usize) {
+        return false;
+    }
+
+    // It's okay.
+    true
+}
+
 // Main function to load the ELF binary
 #[allow(unused)]
 fn load_elf(elf_data: &[u8]) -> Result<(), ElfError> {
@@ -122,7 +180,7 @@ fn load_elf(elf_data: &[u8]) -> Result<(), ElfError> {
                 .cast::<Elf32Phdr>()
         };
 
-        if ph.p_type == SegmentType::Load as u32 {
+        if (ph.p_type == SegmentType::Load as u32) && validate_segment(ph, elf_data) {
             let vm_start = ph.p_vaddr as usize;
             let vm_end = vm_start + ph.p_memsz as usize;
 
--- a/kernel/src/user_program/elf_loader.rs
+++ b/kernel/src/user_program/elf_loader.rs
@@ -134,8 +134,15 @@ fn validate_segment(phdr: &Elf32Phdr, file_data: &[u8]) -> Result<(), ElfSegment
         return Err(ElfSegmentError::DifferentPageOffset);
     }
 
-    // p_offset must point within FILE.
-    if phdr.p_offset as usize > file_data.len() {
+    // The range must point within FILE.
+    // and the region cannot "wrap around" across the kernel virtual
+    // address space.
+    if (phdr
+        .p_offset
+        .checked_add(phdr.p_filesz)
+        .ok_or(ElfSegmentError::VMRegionWrapAround)? as usize)
+        > file_data.len()
+    {
         return Err(ElfSegmentError::OffsetOutOfRange);
     }
 
--- a/kernel/src/user_program/elf_loader.rs
+++ b/kernel/src/user_program/elf_loader.rs
@@ -64,10 +64,9 @@ pub enum ElfError {
 #[derive(Debug)]
 pub enum ElfSegmentError {
     DifferentPageOffset,
-    OffsetOutOfRange,
+    ContentsOutOfRange,
     MemSizeLesserThanFileSize,
     VMRegionOutOfRange,
-    VMRegionWrapAround,
     PageZeroMapping,
     // Additional error types as needed
 }
@@ -140,10 +139,10 @@ fn validate_segment(phdr: &Elf32Phdr, file_data: &[u8]) -> Result<(), ElfSegment
     if (phdr
         .p_offset
         .checked_add(phdr.p_filesz)
-        .ok_or(ElfSegmentError::VMRegionWrapAround)? as usize)
+        .ok_or(ElfSegmentError::ContentsOutOfRange)? as usize)
         > file_data.len()
     {
-        return Err(ElfSegmentError::OffsetOutOfRange);
+        return Err(ElfSegmentError::ContentsOutOfRange);
     }
 
     // p_memsz must be at least as big as p_filesz.
@@ -162,7 +161,7 @@ fn validate_segment(phdr: &Elf32Phdr, file_data: &[u8]) -> Result<(), ElfSegment
     if !is_user_vaddr(
         phdr.p_vaddr
             .checked_add(phdr.p_memsz)
-            .ok_or(ElfSegmentError::VMRegionWrapAround)?
+            .ok_or(ElfSegmentError::ContentsOutOfRange)?
             .saturating_sub(1) as *const (),
     ) {
         return Err(ElfSegmentError::VMRegionOutOfRange);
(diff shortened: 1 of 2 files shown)


serenayu11:
Thanks Ray for making the change :)

kernel/src/user_program/elf_loader.rs
```diff
@@ -122,7 +193,7 @@ fn load_elf(elf_data: &[u8]) -> Result<(), ElfError> {
                 .cast::<Elf32Phdr>()
         };
 
-        if ph.p_type == SegmentType::Load as u32 {
+        if (ph.p_type == SegmentType::Load as u32) && validate_segment(ph, elf_data).is_ok() {
```
mtoohey31, line 196: Using `is_ok` here still just skips the segment if validation fails. I'd recommend something like this:

```diff
diff --git a/kernel/src/user_program/elf_loader.rs b/kernel/src/user_program/elf_loader.rs
index dcd9f2e..b8343f1 100644
--- a/kernel/src/user_program/elf_loader.rs
+++ b/kernel/src/user_program/elf_loader.rs
@@ -56,6 +56,7 @@ pub enum ElfError {
     UnsupportedVersion,
     UnsupportedType,
     UnsupportedMachine,
+    SegmentError(ElfSegmentError),
     // Additional error types as needed
 }
 
@@ -193,7 +194,9 @@ fn load_elf(elf_data: &[u8]) -> Result<(), ElfError> {
                 .cast::<Elf32Phdr>()
         };
 
-        if (ph.p_type == SegmentType::Load as u32) && validate_segment(ph, elf_data).is_ok() {
+        if ph.p_type == SegmentType::Load as u32 {
+            validate_segment(ph, elf_data).map_err(ElfError::SegmentError)?;
+
             let vm_start = ph.p_vaddr as usize;
             let vm_end = vm_start + ph.p_memsz as usize;
```

kernel/src/user_program/elf_loader.rs
```diff
@@ -102,6 +118,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
...
+#[allow(unused)]
+fn is_user_vaddr(vaddr: *const ()) -> bool {
+    unsafe { vaddr < PHYS_BASE }
+}
+
+fn validate_segment(phdr: &Elf32Phdr, file_data: &[u8]) -> Result<(), ElfSegmentError> {
+    // p_offset and p_vaddr must have the same page offset.
+    if (phdr.p_offset as usize & PGMASK) != (phdr.p_vaddr as usize & PGMASK) {
+        return Err(ElfSegmentError::DifferentPageOffset);
+    }
```
mtoohey31, line 135: I'm not clear on why we need to ensure that this is the case. According to the [ELF spec](https://refspecs.linuxfoundation.org/elf/elf.pdf), `p_offset` is the offset within `file_data` of the data to be mapped at the `p_vaddr` virtual address, so I don't see why they'd have to be the same. Is there a reason I've missed?

kernel/src/user_program/elf_loader.rs
```diff
@@ -102,6 +118,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
...
+
+    // p_offset must point within FILE.
+    if phdr.p_offset as usize > file_data.len() {
```
mtoohey31, line 138:
```suggestion
    // The range must be within FILE.
    if (phdr.p_offset + phdr.p_filesz) as usize > file_data.len() {
```

I think we also need to ensure that `p_filesz` doesn't cause the end of the range to extend beyond the end of the data.

kernel/src/user_program/elf_loader.rs
```diff
@@ -102,6 +118,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
...
+
+    // p_memsz must be at least as big as p_filesz.
+    if phdr.p_memsz < phdr.p_filesz {
+        return Err(ElfSegmentError::MemSizeLesserThanFileSize);
+    }
+
+    // The segment must not be empty.
+    if phdr.p_memsz == 0 {
+        return Err(ElfSegmentError::EmptyMemSize);
+    }
```
mtoohey31, line 150: The [ELF spec](https://refspecs.linuxfoundation.org/elf/elf.pdf) says on page 2-2 that `p_memsz` "may be zero".

kernel/src/user_program/elf_loader.rs
```diff
@@ -102,6 +118,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
...
+    if !is_user_vaddr((phdr.p_vaddr as usize + phdr.p_memsz as usize) as *const ()) {
+        return Err(ElfSegmentError::VMRegionOutOfRange);
+    }
+
+    // The region cannot "wrap around" across the kernel virtual
+    // address space.
+    if (phdr.p_vaddr as usize + phdr.p_memsz as usize) < (phdr.p_vaddr as usize) {
+        return Err(ElfSegmentError::VMRegionWrapAround);
+    }
```
mtoohey31, line 165: This check appears to rely on C-like integer overflow wrapping behaviour. In Rust, the program will panic on integer overflow in debug builds. I'd suggest removing this check, and swapping the one above for something like:

```rust
    if !is_user_vaddr(
        phdr.p_vaddr
            .checked_add(phdr.p_memsz)
            .ok_or(ElfSegmentError::VMRegionWrapAround)?
            .saturating_sub(1) as *const (),
    ) {
        return Err(ElfSegmentError::VMRegionOutOfRange);
    }
```

(The `.saturating_sub(1)` will also handle the off-by-one that I mentioned in the previous comment. If it turns out I'm wrong about there being an issue there though you can just remove that method call and the rest should work.)

kernel/src/user_program/elf_loader.rs
```diff
@@ -102,6 +118,61 @@ fn verify_elf_header(header: &Elf32Ehdr) -> Result<(), ElfError> {
...
+    if phdr.p_memsz == 0 {
+        return Err(ElfSegmentError::EmptyMemSize);
+    }
+
+    // The virtual memory region must both start and end within the
+    // user address space range.
+    if !is_user_vaddr(phdr.p_vaddr as *const ()) {
+        return Err(ElfSegmentError::VMRegionOutOfRange);
+    }
+    if !is_user_vaddr((phdr.p_vaddr as usize + phdr.p_memsz as usize) as *const ()) {
```
mtoohey31, line 157: I may be wrong, but I suspect there is an off-by-one error in this check. Since `p_vaddr` is the start address and `p_memsz` is the length, `p_vaddr + p_memsz` is the end of the non-inclusive mapped range, meaning it isn't mapped itself, right? I think we only need to check that `p_vaddr + p_memsz - 1` isn't a user virtual address. You can do this using `.saturating_sub(1)` to ensure you don't overflow if both are zero for some reason.

mtoohey31 (changes requested):
Thanks for making these changes! Looks like we're on the right track, I just have a couple questions about some of the checks.

kernel/src/user_program/elf_loader.rs
serenayu11: https://stackoverflow.com/questions/52533193/how-to-understand-the-difference-between-offset-and-viraddr-in-program-headers-i
From this discussion they commented that "What is guaranteed is that .p_offset % pagesize == .p_vaddr % pagesize (otherwise mmap will become impossible)."

kernel/src/user_program/elf_loader.rs
serenayu11: Yes, but, I think if this is the case we don't need to load this segment into memory?

kernel/src/user_program/elf_loader.rs
serenayu11: I see!

kernel/src/user_program/elf_loader.rs
serenayu11: Then this should not be treated as an error, and continue to load rest of the segments

kernel/src/user_program/elf_loader.rs
mtoohey31: The condition in your comment looks about right, but the one in the code here is a different one. If we want to check for the condition in the comment then I think we should use something like: `p_align == 0 || p_align == 1 || (p_offset % p_align == p_vaddr % p_align)` (see the end of page 2-2 of the ELF spec for explanation of the first two or cases).

kernel/src/user_program/elf_loader.rs
mtoohey31: Yeah this shouldn't be treated as an error. Your call on how you want to change things, there are two possibilities I can think of:
1. Change the return type to `Result<bool, ...>` and use the boolean to indicate whether to skip the segment. If you decided to go with this, then this check should become the last one, cause we should produce an error if there are any other issues with the segment aside from it being 0-length.
2. Just remove the check here and leave it to the code that processes the segments to detect this.

kernel/src/user_program/elf_loader.rs
serenayu11: Resolved in commit 837ff0292a6c0e5e99fc95455e29f5e93a3a1eb9

kernel/src/user_program/elf_loader.rs
serenayu11: Resolved in commit 1219e1f728a9c9d5632ff6a28a6f698bffef7fe2

kernel/src/user_program/elf_loader.rs
serenayu11: I think you are right, I don't know why Pintos doesn't consider this.

serenayu11: Resolved in commit 84694b6c46f2b614d8386ef1b017df08a9335dbc

kernel/src/user_program/elf_loader.rs
serenayu11: Resolved in commit 709bf1472fddc01a89b95cd5acbde9c468000743

kernel/src/user_program/elf_loader.rs
serenayu11: Changed in commit e05b3a76f46eedd171bb8c2bba05a9b56ab9b4dc, could you check if it is good? thanks.

serenayu11: I am thinking to change the naming of ELFSegmentError:VmRegionWrapAround to ELFSegmentError:SegmentOverflow, do you think this would be more accurate?

kernel/src/user_program/elf_loader.rs
serenayu11: I thought both calculations were used to ensure that p_offset and p_vaddr are aligned to the same position within a page? Just one using bitwise operation and one using %? Sorry, correct me if I am wrong.

mtoohey31: Oh I see, you're trying to use bitwise and instead of remainder? In that case I think `PGMASK` might be the wrong value to use, cause in binary its current value is `10111111111111111111111111111111`, which won't give us the value modulo the page size. And also, we shouldn't be basing this check on our own page size constant anyway, cause as the section of the ELF spec that I linked mentions, the only guarantee is that the two are equal modulo `p_align` if it's not 0 or 1, and `p_align` might not correspond to our page size.

kernel/src/user_program/elf_loader.rs
mtoohey31: Good idea with the use of `checked_add`. What about renaming `OffsetOutOfRange` to `ContentsOutOfRange` and using it both when overflow occurs, and when the result doesn't overflow but is bigger than `file_data.len()`?

serenayu11: Sure

kernel/src/user_program/elf_loader.rs
serenayu11: image
Seems like it is 111111111111? Binary 4095

kernel/src/user_program/elf_loader.rs
serenayu11: Changed in commit 6abe23e161957c47adfdb991cf3bf299b4b74471

kernel/src/user_program/elf_loader.rs
mtoohey31: Yes, but we can't be sure 4096 is the correct alignment to expect. As per the second half of my comment above, that's what `p_align` is for.

kernel/src/user_program/elf_loader.rs
mtoohey31: Oh sorry, forgot that `VMRegionWrapAround` was used below too. I don't think we should use `ContentsOutOfRange` there cause that doesn't really describe what went wrong in that case. Could we maybe leave `VMRegionWrapAround` or change it to `VMRegionOutOfRange` on line 164? Sorry about the back and forth...

serenayu11: @mtoohey31 Changed in commit 90e94ff4881bca3ef9f629a8c70d286a75157db7 :)

kernel/src/user_program/elf_loader.rs
serenayu11: @mtoohey31 Change in commit 8f52249656a850766760fd4d906555176832d658

mtoohey31 (approved):
Thanks for making the changes and adding comments. Looks good now!

PR #21, opened 2024-03-26
documentation using mdbook

mtoohey31:
This pull request adds documentation, which can be built into a website using [mdBook](https://github.com/rust-lang/mdBook). It also makes some changes to the Docker build process.


docs/bootloader-to-main.md
DerekCresswell, line 1: Love this!

mtoohey31:
Ok this should be ready now. I've added workflow steps for deploying the docs. They're skipped when the workflow is run for this pull request cause we only want things to be deployed on main (but I've checked that they work in [this workflow run](https://github.com/KidneyOS/KidneyOS/actions/runs/8650535070) from before I added the conditional skipping).

ff-zhang:
We are planning to merge in the docs in their unfinished state. This is done to speed up onboarding new contributors and provide more guidance, however limited. As various components of the kernel is completed, the associated portion of the docs will be filled in.

docs/building/macos.md
```diff
@@ -0,0 +1,64 @@
+# macOS (ARM)
```
1whatleytay, line 1: Great!

docs/building/from-source.md
```diff
@@ -0,0 +1,51 @@
+# From Source
+
+This section describes how to build KidneyOS by installing dependencies using a package manager and building it from source.
+
+> ❗ This method is only supported on `x86_64-linux` and `aarch64-linux` hosts, or in WSL. For building on macOS (ARM), see [here](./macos.md).
```
1whatleytay, line 5: Nit: Might be worth linking to macOS from here since these rust target steps also need to be followed, or maybe mention in the macOS docs that these are additional steps?

1whatleytay (approved):

PR #22, opened 2024-03-29
Thread Reaping

DerekCresswell:


DerekCresswell:
This is failing from underflow on the interrupt lock.

DerekCresswell:
After getting the interrupts into a good state, I find this getting a triple fault when returning from the context switch into the idle function.

I thought maybe it could be to do with returning the Option from ThreadFunctions, as Rust reports this is not FFI-safe. However this did not alleviate the issue.

This error also occurred in the interrupts before #18 .  The issue was with the stack pointer being decremented / incremented by non-aligned amounts.

DerekCresswell:
Added here (though not dependent on the other code in the PR) is updated documentation for the threading Readme. I hope it, combined with any code comments, should aide in getting developers up to speed on KidneyOS's threading system.

ff-zhang:
Changes incorporated into main already. Opened a new issue for current thread cleanup bugs.

PR #23, opened 2024-03-29
Adding Interrupt lock that restores to the previous interrupt level

serenayu11:
I have added a different type of interrupt lock that uses the method of restoring to the previous interrupt level. It is called InterruptLock, the counter interrupt lock implementation remains in the file but has been renamed to InterruptCounterLock.

### DIFF
--- a/kernel/src/sync.rs
+++ b/kernel/src/sync.rs
@@ -68,25 +68,20 @@ impl<T> Drop for SpinLockGuard<'_, T> {
 
 static INTR_DISABLE_COUNT: AtomicUsize = AtomicUsize::new(0);
 
-// This function disables interrupt, and increments the interrupt count when called
+// This function disables interrupt
 #[allow(unused)]
 pub fn intr_disable() {
-    // Increment the disable count atomically
-    INTR_DISABLE_COUNT.fetch_add(1, Ordering::SeqCst);
     // disable
     unsafe {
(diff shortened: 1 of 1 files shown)


kernel/src/sync.rs
```diff
@@ -127,26 +188,38 @@ unsafe impl<T> Send for InterruptLock<T> {}
...
-        intr_enable();
+        let previous = self.intr_level.load(Ordering::SeqCst);
+        let previous_intr_level = match previous {
+            0 => IntrLevel::IntrOn,
+            1 => IntrLevel::IntrOff,
+            _ => panic!("Unexpected value stored in intr_level"),
+        };
+        if previous_intr_level == IntrLevel::IntrOn {
+            intr_enable();
+        }
```
DerekCresswell, line 222: I'm not sure this will let us achieve what we want. I don't believe that you can store the previous level in the lock itself.

I believe (and correct me if I'm wrong), we need lock to return the previous level and unlock to accept the new level.
That previous value would need to be on the stack for whoever called it.

DerekCresswell (changes requested):

DerekCresswell:
I pulled a different version of this out in PR #22 

From our meeting discussions it is probably easiest to use my 'function' version over these lock version. Unfortunately, this is less 'rusty' than the locks.

PR #24, opened 2024-04-02
Add code coverage targets

mtoohey31:
This merge request adds targets for checking code coverage to our Makefile. It doesn't enforce any specific level of code coverage. I don't think aiming for a certain % makes sense since there are lots of things that we just can't do in tests (for example, anything that requires the CPU to be in kernel mode). I'll add instructions about how to use this to our mdbook documentation once either this or #21 is merged.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,7 @@ jobs:
       run: nix --extra-experimental-features nix-command --extra-experimental-features flakes develop ./nix#build -i -k CARGO_TERM_COLOR -c sh -c 'find -name "*.rs" -print0 | xargs -0 rustfmt --quiet --check --'
     - name: Run tests
       if: ${{ !cancelled() && steps.cache-nix.outcome == 'success' }}
-      run: nix --extra-experimental-features nix-command --extra-experimental-features flakes develop ./nix#build -i -k CARGO_TERM_COLOR -c make test
+      run: nix --extra-experimental-features nix-command --extra-experimental-features flakes develop ./nix#build -i -k CARGO_TERM_COLOR -c make print-coverage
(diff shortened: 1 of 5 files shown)


DerekCresswell:
> > My only question, do we want all of our CI checks as part of one job? Usually we would have one for coverage, one for tests, one for build...
> 
> Now that you mention it, I do remember seeing lots of other projects do that lol. Do you know what the reasoning behind that is?
> 
> If it's to make sure we still see the results of each check if one of them fails, I managed to get that working with `if` checks in the steps (see the most recent failure [here](https://github.com/KidneyOS/KidneyOS/actions/runs/8633230781/job/23665795107) for an example). Maybe there are other benefits though?

Oh the if's do do that. I think the only thing I'd say is that on the PR page it's all one. But it's a real nitpick!

mtoohey31:
> Oh the if's do do that. I think the only thing I'd say is that on the PR page it's all one. But it's a real nitpick!

Ah true, it is nicer to have them split up... the reason I'm a little reluctant about changing it is this failure: https://github.com/KidneyOS/KidneyOS/actions/runs/8531010432/job/23369861402. It looks like the error happened cause the caching action was running simultaneously in two jobs after I made changes to the flake. I probably just need to clean up the caching action, but that'll take a bit of work, and if we change things without doing that I suspect we'll have to deal with the annoyance of all but one jobs failing each time we change the flake...

Is leaving a TODO to split things up a decent compromise?

DerekCresswell:
I'm fine for you to leave it as is : )

Been a while since I wrote an action, so I'd need to look up the specifics. But we can have a set of dependencies so that the jobs always run in a specified order. But I see this is already written as 'steps' so it's really got no reason to change.

PR #25, opened 2024-04-10
To User Space and Back

mtoohey31:
This pull request will implement support for starting programs in user space, then getting back to kernel mode by executing `int 0x80`.


mtoohey31:
So I had to make some changes to the `ThreadControlBlock` to make this work... What I've pushed now does get to userspace and back (and back to userspace again, at which point it hangs in the test program that loops if the unimplemented exit syscall doesn't stop the program). There are still a bunch of TODOs but I don't want to fix all of them right now cause I anticipate that just leading to more conflicts with #22, plus a lot of them are good small tasks that new team members could work on to get started.

parth2324 (approved):
I went over all the changes, and below is a long summary of what I read and approved:


.githooks/pre-commit:
-> makes programs.mk, which builds programs/exit/exit from programs/exit/exit.S

programs/exit/exit.S:
-> Boch's break instruction, then exit syscall with code 0, (eax: 0x1, ebx: 0x0) via int 0x80, then infinite loop if returned.

.github/workflows/ci.yaml:
-> CI now also tries making programs.mk under "build test programs", and "check clippy" shifted down in order, right after "build test programs".

Cargo.lock:
-> includes "lazy_static" for kidneyOS, implying static variables can be evaluated at runtime, with std env not necessarily required.

Makefile:
-> CARGO_TARGET_DIR and PROFILE changed to conditional assignment, adjusted to include "exit" from programs.mk.

build-support/i686.ld:
-> OFFSET: 0xC => 0x8

kernel/src/interrupt_descriptor_table.rs:
-> added syscall handler connection on 0x80.

kernel/src/main.rs:
-> loads gobal_descriptor_table, has "exit" object file as INIT.

kernel/src/mem/mod.rs:
-> defines frame allocator wrapper and implements new_in(), alloc(), dealloc(); Kernel allocator state modified to use frame allocator wrapper, getting rid of frames_base field, and converting Range<usize> => NonNull<[u8]> in subblock_allocators field.
-> bootstrap allocation initialization modified to be consistent with new structures, and public unsafe frame_alloc(), frame_dealloc() implemented for kernel allocator. The return type for alloc() switches from Optional<Range<usize>> to Result<NonNull<[u8]>, AllocError>.

kernel/src/paging.rs:
-> rearrangement for cleaner loading code / setting up default, basic behavior when creating page manager.

kernel/src/threading/context_switch.rs:
-> appended loading page manager when switching threads, switching stacks during context switch requires kernel stack pointer offset in modified implementation.

kernel/src/threading/mod.rs:
-> starting thread system after initialization requires a page manager for kernel; test_func, test_func_2 replaced with init_elf, pushed to the scheduler.

kernel/src/threading/thread_control_block.rs:
-> NEXT_UNRESERVED_TID no longer mutable, making allocate_tid a safe operation; Defined thread stack number of frames and sizes for both user and kernel, also defining the bottom of user stack in VM: 0x1.
-> TCB now replaces a simple stack pointer and stack pointer bottom with a seperate kernel stack pointer, kernel stack (to free later), eip to store next user instruction, esp to store user stack pointer, and the user stack itself (again to free later); Also now includes a page manager.
-> TCB create() now accepts ELF data as a reference to u8 array. It parses the ELF data to get entrypoint and vm_areas, and the KERNEL_ALLOCATOR is used to allocate appropriate numer of frames for each area, mapping them to the virtual address assigned by ELF header. Then write from the ELF data for each area (given by offset) to the physical address, writing 0 for any space after length until the end of the last frame. The function then allocates a kernel stack, writing 0 everywhere. We currently don't check whether the stack section is already defined in the ELF data, and give the thread a user stack with hardcoded user stack bottom VM address. Finally, a switch threads context is added to the stack, and the status is set to ready before returning the new thread tcb.
-> Creating a kernel thread now requires a page manager, and the tcb fields are set to NonNull::dangling() so that they will be set in the context switch immediately following the creation. has_stack_space() and shift_stack_pointer_down() are modified to work on tcb's kernel stack.

kernel/src/threading/thread_functions.rs:
-> run_thread() now accepts only 'to' and 'from' thread TCBs, in which the status of 'to' thread is set to running, TASK_STATE_SEGEMENT.esp0 is set to point to the kernel stack, RUNNING_THREAD is set to the thread, and the 'from' thread is pushed to the scheduler. After enabling interrupts, finally uses iret method to enter user mode and execute the 'to' thread. This involves setting the data segment (DS), extra segement (ES), thread local storeage (FS), processor data area (GS), with code segment (CS) and stack segment (SS) set by iret. It also prepares the stack frame for iret, pushing in order: stack segment, esp, flags, code segment, and eip.

kernel/src/user_program/elf_loader.rs:
-> load_elf() replaced with parse_elf(), which also now returns the entry address and VM area instead of unit upon success. The program section offset from program header is now also passed to VMAreaStruct.

kernel/src/user_program/syscall.rs:
-> made handler(), needs implementation for different syscall numbers.

kernel/src/user_program/virtual_memory_area.rs:
-> VM area struct now hold offset as well, and requires it for initialization.

shared/src/global_descriptor_table.rs:
-> In Segment Descriptor structure, base_low replaced with base, GDT_LEN 5 => 6 since added TSS (Task State Segment), GDT no longer const but static mutable.
-> Made public const KERNEL_CODE_SELECTOR, KERNEL_DATA_SELECTOR, USER_CODE_SELECTOR, USER_DATA_SELECTOR, and just const local TSS_INDEX = 5, TSS_SELECTOR.
-> In load(), GDT[TSS_INDEX] is set base to segment::TASK_STATE_SEGMENT, and lgdt (load global descriptor table), ltr (load task register) used to fill up GDT_DESCRIPTOR, TSS_SELECTOR respectively. KERNEL_CODE_SELECTOR & KERNEL_DATA_SELECTOR are also set appropriately.

shared/src/mem/mod.rs:
-> added VM main_stack_top(), still using to_virt!() macro, and redefined OFFSET 0xC => 0x8 (virtual addresses above or at OFFSET are kernel addresses), and also defined bootstrap allocator size as 8 MB (recommended to be re-evaluated).

shared/src/paging.rs:
-> implemented page_table() for PageDirectory to get page table reference from self, virt_parts() helper to get page directory and page table indices (VirtualAddress struct moved inside the function), and map() modified to use the two implemented functions.
-> changed behavior for map() s.t., 'write' and 'user' bools indicate whether to enable read_write and user_supervisor respectively for page directory, and thus, if true, then set to enabled, otherwise may or may not be enabled. Earlier, if true, then enabled, otherwise disabled.
-> huge_map() now enforces that physical address is page-aligned and not necessarily huge-page-aligned. Also modified to utilize virt_parts().
-> implemented Clone for PageManager<A: Allocator + Copy>, unsafe 'root' clone to pointer from attempting to allocate PAGE_DIRECTORY_LAYOUT.

shared/src/segment.rs:
-> defined SegmentSelector public bitfield structure.

shared/src/task_state_segment.rs:
-> defined TSS public struct, and the public static mutable TASK_STATE_SEGMENT as an all 0 TSS struct, but ss0 pointing to KERNEL_DATA_SELECTOR, and iopb (IO port bitmap) to the size of TSS struct.


CI is happy as well, thus we can merge with main and starting filling the TODOs.

PR #31, opened 2024-06-03
Implment syscalls

ff-zhang:
This PR implements the syscalls left as TODOs in the main branch. It relies on the changes in https://github.com/KidneyOS/KidneyOS/pull/22 to handle scheduling operations.


ff-zhang:
Stuck on an issue where, during the first context switch from the kernel thread to the user thread, the `PageManager` complains it is dropped while it is still loaded. When ignored, there is a separate issue with an internal inconsistency within the kernel allocator.

kernel/src/threading/thread_control_block.rs
```diff
@@ -27,12 +27,12 @@ static NEXT_UNRESERVED_TID: AtomicU16 = AtomicU16::new(0);
 // Windows: https://techcommunity.microsoft.com/t5/windows-blog-archive/pushing-the-limits-of-windows-processes-and-threads/ba-p/723824
 pub const KERNEL_THREAD_STACK_FRAMES: usize = 2;
 const KERNEL_THREAD_STACK_SIZE: usize = KERNEL_THREAD_STACK_FRAMES * PAGE_FRAME_SIZE;
-pub const USER_THREAD_STACK_FRAMES: usize = 4 * 1024;
+pub const USER_THREAD_STACK_FRAMES: usize = 256;
```
ff-zhang, line 30: In order to be consistent with the rest of the the consts, it might be better to do something like `KB >> 2`

kernel/src/user_program/syscall.rs
```diff
@@ -1,30 +1,129 @@
...
+            let parent_tid = running_tcb.tid;
+
+            let child_tcb = (**running_tcb).clone();
+            let child_tid = child_tcb.tid as usize;
+
+            if parent_tid == running_tcb.tid {
+                child_tid
+            } else {
+                // Still gettng an error that the page table is being dropped while loaded here
+                println!("{}", child_tcb.page_manager.is_loaded());
```
ff-zhang, line 37: Do you know whether the child or parent is causing this?

kernel/src/user_program/syscall.rs
```diff
@@ -1,30 +1,129 @@
...
+            //             .as_mut()
+            //             .expect("Scheduler not set up!")
+            //             .remove(running_tcb.tid)
+            //     };
+
+            //     let waited_tcb = unsafe {};
+            // }
+        }
+        SYS_EXECVE => {
+            // todo!("exec syscall")
```
ff-zhang, line 82: This should be left as a `todo!` instead of returning `1`

ff-zhang (commented):
Ideally, we would have a separate PR for the FIFO scheduler. It seems small so it might be fine to leave here though

kernel/src/user_program/syscall.rs
```diff
@@ -56,31 +58,34 @@ pub extern "C" fn handler(syscall_number: usize, arg0: usize, arg1: usize, arg2:
             2048
         }
         SYS_WAITPID => {
-            todo!("waitpid not implemented");
-            // println!("Starting wait syscall");
+            // todo!("waitpid not implemented");
+            println!("Starting wait syscall");
```
ff-zhang, line 62: I assume this is for logging but this and the `todo!` should be removed in the final version

ff-zhang:
Also, we will need to rebase against `main` again because of how syscalls will need be to be registered for the compilation of `C` and `Rust` binaries.

kernel/src/user_program/syscall.rs
mckeenicholas: Sorry I missed this comment. It's the child that is causing it.

PR #32, opened 2024-06-19
Add TicketMutex

steven-mathew:
`TicketMutex` grants the lock to the first thread that requested it. Also added `Mutex` providing an abstraction over the type of the mutex we use, which is enabled through a cfg directive.


ff-zhang (approved):
LGTM! We might need to, in the future, handle the case where it runs out of tickets though.

steven-mathew:
> LGTM! We might need to, in the future, handle the case where it runs out of tickets though.

This won't be an issue because we won't have 2^32 (give or take a few) waiters at a time.

ff-zhang:
Right, but we don't reclaim the tickets in anyway so, theoretically, we could run out eventually.

PR #33, opened 2024-07-06
Implement PCB/TCB Split

ff-zhang:
Spins off PCB functionality from the TCB implementation. This is also used to enable us to differentiate between user and kernel threads during context switches. This is preparatory work for our syscall implementation (https://github.com/KidneyOS/KidneyOS/pull/31).


kernel/src/threading/thread_control_block.rs
```diff
@@ -151,51 +251,33 @@ impl ThreadControlBlock {
...
-    pub unsafe fn create_kernel_thread(page_manager: PageManager) -> Self {
+    pub unsafe fn new_kernel_thread(page_manager: PageManager) -> Self {
         ThreadControlBlock {
             kernel_stack_pointer: NonNull::dangling(), // This will be set in the context switch immediately following.
             kernel_stack: NonNull::dangling(),
             eip: NonNull::dangling(),
             esp: NonNull::dangling(),
             user_stack: NonNull::dangling(),
             tid: allocate_tid(),
+            pid: allocate_pid(),
```
steven-mathew, line 280: I believe we said we should give the kernel thread a pid of 0.

kernel/src/threading/thread_functions.rs
```diff
@@ -52,32 +52,53 @@ unsafe extern "C" fn run_thread(
...
     // Our scheduler will operate without interrupts.
     // Every new thread should start with them enabled.
+    outb(0x21, 0xfd);
+    outb(0xa1, 0xff);
     intr_enable();
 
-    // https://wiki.osdev.org/Getting_to_Ring_3#iret_method
-    // https://web.archive.org/web/20160326062442/http://jamesmolloy.co.uk/tutorial_html/10.-User%20Mode.html
+    // Kernel threads have no associated PCB, denoted by its PID being 0
+    if pid == 0 {
```
steven-mathew, line 60: Ah, yep we should! We can merge it as is though, I've made the fix in my upcoming PR.

steven-mathew (approved):

kernel/src/threading/thread_control_block.rs
ff-zhang: Since this is the first PID allocated, it should be given a value of `0` here.

PR #34, opened 2024-08-03
Enables timer interrupts and adds a system clock

ff-zhang:
Enables timer interrupts by programming the PICs and PIT when initializing the OS. Uses this to implement a coarse system clock and a simple `sleep` (kernel) function.


steven-mathew:
Perfect! Also, thanks for adding exit code and reaping to this PR as well.

steven-mathew (approved):
Perfect! Also, thanks for adding exit code and reaping to this PR as well. There are a few minor nitpicks, but I will address them in my PR with the preemption guard.

PR #35, opened 2024-08-06
Add irq safe mutex and interrupts guard

steven-mathew:
Adds `MutexIrq` and an interrupt guard so that the user doesn't have to disable the held guard and it'll drop when it falls out of scope.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,12 @@
 result
 /.direnv
 
+/target/
 /build/
 bx_enh_dbg.ini
 *.core
 /local.mk
 *.profraw
 rustc-ice-*.txt
+kidneyos-builder.tar.gz
+.DS_Store
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -7,6 +7,7 @@
 #![feature(slice_ptr_get)]
 #![cfg_attr(target_os = "none", no_std)]
 #![cfg_attr(not(test), no_main)]
+#![feature(negative_impls)]
 
 mod interrupt_descriptor_table;
 mod mem;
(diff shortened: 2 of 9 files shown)


ff-zhang:
Can this PR include the guard for the `SYS_CLOCK` in `timer.rs`

ff-zhang (approved):
One small bug to fix but everything else looks good to me.

kernel/src/sync/irqsafe.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct MutexGuardIrqSafe<'a, T: ?Sized + 'a> {
+    guard: MutexGuard<'a, T>,
+    _guard: InterruptsGuard,
+}
+
+// Same unsafe impls as `std::sync::MutexIrqSafe`
+unsafe impl<T: ?Sized + Send> Sync for MutexIrqSafe<T> {}
+unsafe impl<T: ?Sized + Send> Send for MutexIrqSafe<T> {}
+
+impl<T> MutexIrqSafe<T> {
```
ff-zhang, line 23: Should this be called `MutexIRQ` instead of `MutexIRQSafe`?

steven-mathew: Sure!

kernel/src/threading/mod.rs
```diff
@@ -57,6 +57,12 @@ pub fn thread_system_start(kernel_page_manager: PageManager, init_elf: &[u8]) ->
             .push(Box::new(user_tcb));
     }
 
+    // Enable preemptive scheduling.
```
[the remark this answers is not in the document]
steven-mathew: Done.

steven-mathew:
> Can this PR include also add the guard to the `SYS_CLOCK` in `timer.rs`?

Working on it!

PR #42, opened 2024-09-17
VFS layer

pommicket:
Here is an initial version of the VFS layer, together with a "reference implementation" of an in-memory filesystem. The key thing here is the `FileSystem` trait which we can implement for each supported filesystem.  I've tried to follow the FUSE API quite closely, with some simplifications. Let me know what you think!

- Right now there is no mode/owner/group for files. I'm not sure whether we're planning on having multi-user support, but it shouldn't be too hard to add this if we are.
- Also, access/create/modify times can be added in the future.
- ~~It's quite difficult to handle the `offset` argument to `readdir` correctly (in fact, most FUSE filesystems I found online just ignore it), i.e. in such a way that unrelated files are never repeated/skipped if a directory is modified between calls. But I'm not sure if there's a better alternative, assuming we want a POSIX `readdir`-like userspace API (one option is to remove `offset` and just read all the entries into memory in the kernel whenever a directory is opened).~~ (readdir now works in a different way that avoids this issue — we will just read all directory entries into memory whenever a directory is opened; it's probably more efficient to do that anyways)
- Right now filenames are required to be valid UTF-8, which isn't required on Linux. I think this is a good idea in theory, but if we want to be entirely compatible with existing filesystems, we could change this.

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -6,7 +6,6 @@
 #![feature(non_null_convenience)]
 #![feature(offset_of)]
 #![feature(slice_ptr_get)]
-#![feature(error_in_core)] // stabilized in 1.81.0
 #![cfg_attr(target_os = "none", no_std)]
 #![cfg_attr(not(test), no_main)]
 #![feature(negative_impls)]
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -6,6 +6,7 @@
 #![feature(non_null_convenience)]
 #![feature(offset_of)]
 #![feature(slice_ptr_get)]
+#![feature(error_in_core)] // stabilized in 1.81.0
 #![cfg_attr(target_os = "none", no_std)]
 #![cfg_attr(not(test), no_main)]
 #![feature(negative_impls)]
@@ -19,6 +20,7 @@ mod sync;
 mod threading;
 mod timer;
 mod user_program;
+pub mod vfs;
 
 extern crate alloc;
 
--- a/kernel/src/vfs/mod.rs
+++ b/kernel/src/vfs/mod.rs
@@ -1,7 +1,8 @@
+#[cfg(test)]
+pub mod read_only_test;
 pub mod tempfs;
 
-use alloc::borrow::ToOwned;
-use alloc::string::String;
+use alloc::{borrow::Cow, format, string::String, vec::Vec};
 
 pub type INodeNum = u32;
 pub type Path = str;
@@ -12,12 +13,11 @@ pub type OwnedPath = String;
 /// **IMPORTANT**: the kernel must call [`FileSystem::release`]
 /// when it closes its last open file to an inode. Otherwise,
 /// the filesystem will have to keep around the file's data indefinitely!
-pub trait FileHandle: Copy {
-    /// get inode number of this file
+pub trait FileHandle: core::fmt::Debug {
     fn inode(self) -> INodeNum;
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub enum Error {
     /// directory entry not found
     NotFound,
@@ -33,8 +33,16 @@ pub enum Error {
     NotEmpty,
     /// Target destination already exists
     Exists,
+    /// Unsupported operation (e.g. file system does not support symlinks)
+    Unsupported,
     /// Error accessing underlying storage device
-    IO,
+    IO(String),
+}
+
+impl From<crate::block::block_error::BlockError> for Error {
+    fn from(value: crate::block::block_error::BlockError) -> Self {
+        Self::IO(format!("{value}"))
+    }
 }
 
 impl core::fmt::Display for Error {
@@ -47,7 +55,8 @@ impl core::fmt::Display for Error {
             Self::TooManyLinks => write!(f, "too many hard links to file"),
             Self::NotEmpty => write!(f, "directory not empty"),
             Self::Exists => write!(f, "destination already exists"),
-            Self::IO => write!(f, "I/O error"),
+            Self::Unsupported => write!(f, "unsupported operation"),
+            Self::IO(s) => write!(f, "I/O error: {s}"),
         }
     }
 }
@@ -57,6 +66,7 @@ impl core::error::Error for Error {}
 pub type Result<T> = core::result::Result<T, Error>;
 
 /// File or directory information, as returned by stat.
+#[derive(Debug, Clone)]
 pub struct FileInfo {
     /// Whether this is a file, directory, etc.
     pub r#type: INodeType,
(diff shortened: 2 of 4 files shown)


kernel/src/main.rs
```diff
@@ -16,6 +17,7 @@ mod sync;
 mod threading;
 mod timer;
 mod user_program;
+pub mod vfs;
```
1whatleytay, line 20: What is `pub mod` going to do if this is a binary?

1whatleytay:
Cool!

kernel/src/main.rs
pommicket: That was just to suppress unused function warnings (e.g. TempFs isn't used anywhere else in kidneyos currently). I don't think it makes much of a difference to the binary, and it's a bit nicer than `#![allow(dead_code)]`  (which also suppresses warnings for private functions, etc.).

kernel/src/vfs/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+///
+/// **IMPORTANT**: the kernel must call [`FileSystem::release`]
+/// when it closes its last open file to an inode. Otherwise,
+/// the filesystem will have to keep around the file's data indefinitely!
+#[derive(Debug, Clone, Copy)]
+pub struct FileHandle {
+    /// inode number of this file
+    pub inode: INodeNum,
+    /// allows filesystem to store its own metadata about open files
+    pub fs_data: usize,
```
1whatleytay, line 20: Proposal: I don't really like this `usize` if it's intended to be used like a `void *` (e.g. we'd have to cast it to something useful probably, and now we don't get some of the nice reference tracking from Rust). Is the `FileSystem` trait intended to be `dyn`?

Something like an associated type `trait FileSystem { type Handle: SomeHandleTrait; }` would be amazing. This means we can't mix and match say, `FAT16` handles with `PizzaFS` handles + this would dance around all the `MaybeUninit` unsafe magic we'd have to do. Worst case, the user could `Box<dyn SomeHandleTrait>`. Presumably, the `FileHandle` would have to be responsible with owning `fs_data` on the heap anyway, so we really aren't allocating doing any more work here than before.

And if it really has to return some kind of `void *` object, I would almost prefer `core::any::Any` here.

1whatleytay: Or if we don't know how we are going to use `fs_data` and it's not intended to be serialized to disc or something, maybe we should add it later so we can reason a bit better about how we should type it?

kernel/src/vfs/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+/// File or directory information, as returned by stat.
+pub struct FileInfo {
+    /// Whether this is a file, directory, etc.
+    pub r#type: INodeType,
+    /// inode number
+    pub inode: INodeNum,
+    /// Size in bytes
+    pub size: u64,
+    /// Number of hard links
+    pub nlink: u16,
```
1whatleytay, line 71: Seems like this isn't really a big deal (FAT doesn't really seem to even support hard links), but `u16` seems like it can be exhausted really easily? What's `TooManyLinks` for? Is there some kind of abuse we're protecting against?

kernel/src/vfs/mod.rs
pommicket: Yes, I like the idea of an associated type for the file handles — I'll go ahead and change it to that. The fs_data is copied from the FUSE API (fh member of https://libfuse.github.io/doxygen/structfuse__file__info.html), and I assume it has some use there (we might want to track some information about open files in the file system). Right now the FileSystem trait isn't object-safe because of the directory iterator— I'm not sure if there's a good way of dealing with that, but idk if we will actually ever need `dyn FileSystem`.

kernel/src/vfs/mod.rs
pommicket: Well, FWIW `TooManyLinks` exists on linux as `EMLINK`. If we are implementing an existing file system with a low hard link limit, we will need some error to report when we exceed that limit  -- of course, we could just use some more generic error (Error::IO or whatever) but maybe this is a bit nicer idk.

I vaguely remembered 65,536 being the limit for most filesystems, but it looks like it might be possible to exceed that on ext4 with special options- I just checked and nlink_t is actually 8 bytes on x86-64 (so I think we should bump this to u32).

kernel/src/vfs/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Look up file in directory
+    fn lookup(&self, parent: Self::FileHandle, name: &Path) -> Result<INodeNum>;
+    /// Open an existing file/directory/symlink.
+    ///
+    /// If the inode doesn't exist, returns [`Error::NotFound`].
+    fn open(&mut self, inode: INodeNum) -> Result<Self::FileHandle>;
+    /// Create a new file in parent, or open it if it already exists (without truncating).
+    ///
+    /// The kernel must ensure that `parent` is a directory and that `name` is non-empty and doesn't contain `/`
+    fn create(&mut self, parent: Self::FileHandle, name: &Path) -> Result<Self::FileHandle>;
```
1whatleytay, line 140: Beautiful ❤️

pommicket:
FYI: I've added a `SimpleFileSystem` trait which is easier to implement but less flexible than `FileSystem` (any type implementing SimpleFileSystem automatically implements FileSystem) — the main difference is that instead of an associated FileHandle type for open files, `SimpleFileSystem` just uses inode numbers. I think it's still good to have the option of keeping track of extra state for open files (e.g. I was using it in an earlier version of the FAT filesystem to keep track of which disk sector corresponds to the current file offset — they're stored using a linked list in FAT so finding the sector for an offset is O(n). But now we just generate an array of sectors when open() is called since I don't think the overhead is that bad).

kernel/src/vfs/mod.rs
```diff
@@ -0,0 +1,472 @@
...
+        Err(Error::Unsupported)
+    }
+    /// Read the contents of a symbolic link
+    fn readlink(&mut self, link: INodeNum) -> Result<String> {
+        Err(Error::Unsupported)
+    }
+    /// Version of [`SimpleFileSystem::readlink`] that doesn't allocate.
+    ///
+    /// If you implement [`SimpleFileSystem::readlink`], this will be provided automatically.
+    fn readlink_no_alloc<'a>(
```
1whatleytay, line 370: This method doesn't seem that useful to me: like if I don't allocate enough space I have to allocate more in a loop?

Maybe add an offset parameter so I can read 20 bytes at a time.

Also the heapless crate would make a method like this a bit nicer to use [(heapless string)](https://docs.rs/heapless/latest/heapless/struct.String.html). It would mean you would have a compile time fixed size, but you wouldn't need to pass in a buf.

1whatleytay: Heapless might be cursed in the context of a trait but I haven't thought of that too much.

1whatleytay (approved):

kernel/src/vfs/mod.rs
pommicket: The idea here was that the readlink syscall could try with a buffer on the stack, then switch to a heap-allocated buffer if it's too small (this interface lets us e.g. change the threshold for using the heap without changing any of the filesystem implementations). you don't really need to call readlink in a loop because you can use stat to get the size of the link (which won't change btwn stat and readlink if you have an exclusive reference to the fs). It's probably fine to have a small heap allocation for each call to readlink, but at least this leaves us the option to not do that.

1whatleytay: Isn't that what Cow is for?

pommicket: Cow lets you return either a heap allocation or a reference, but not a stack allocation. We could instead use ArrayString from the arrayvec crate if we want the option of returning a heap or a stack allocation. I don't think it makes much of a difference though- we can always add a helper function inside the kernel that returns an ArrayString if we want to.

1whatleytay: Why not a stack allocation? Doesn't seem any more special than a regular reference (why couldn't we just return `buf`).
```
pub enum Cow<'a, B>
where
    B: 'a + ToOwned + ?Sized,
{
    Borrowed(&'a B),
    Owned(<B as ToOwned>::Owned),
}
```

pommicket: Oh I see, you mean still have `buf` as a parameter and return a Cow to either it or a dynamic allocation. I guess we could do that. Ultimately this version mirrors the syscall API more closely (we will presumably have something like `readlink(int fd, char *buf, size_t buf_size)`, since we can't return a dynamic allocation from a syscall) so I think it's a bit better. In the end this function is likely only going to be called in a couple of places in the kernel anyways (namely when resolving paths and when implementing the readlink syscall).

PR #43, opened 2024-09-18
WIP: Add basic interrupt handler

LanceLi1416:
Adds a basic interrupt handler that aligns with pintos/threading/interrupt.

## Usage
```rust
// main.rs
...
let pic: Pic = intr_init()
...
```

Pass `pic` into the `init` functions to register interrupts and their handlers.

> **Note:** currently this interrupt system  does not interact with anything in the code.

The following API's are provided
```rust
/// Returns the current interrupt level.
pic.intr_get_level(&self) -> IntrLevel

/// Enables or disables interrupts as specified by `level'.
///
/// Returns the previous interrupt state.
pic.intr_set_level(&self, level: IntrLevel) -> IntrLevel

/// Enables interrupts and returns the previous interrupt state.
pic.iter_enable(&self) -> IntrLevel

/// Disables interrupts and returns the previous interrupt status
pic.intr_disable(&self) -> IntrLevel

/// Registers external interrupt VEC_NO to invoke HANDLER, which is named NAME for debugging
/// purposes. The handler will execute with interrupts disabled.
pic.intr_register_ext(&mut self, vec_no: u8, handler: IntrHandlerFunc, name: &'static str)

/// Registers internal interrupt `vec_no` to invoke HANDLER, which is named `name` for debugging
/// purposes. The interrupt handler will be invoked with interrupt status `level`.
///
/// The handler will have descriptor privilege level `dpl`, meaning that it can be invoked
/// intentionally when the processor is in the `dpl` or lower-numbered ring.  In practice, `dpl==3`
/// allows user mode to invoke the interrupts and `dpl==0` prevents such invocation. Faults and
/// exceptions that occur in user mode still cause interrupts with `dpl==0` to be invoked. See
/// [IA32-v3a] sections 4.5 "Privilege Levels" and 4.8.1.1 "Accessing Nonconforming Code Segments"
/// for further discussion.
pic.intr_register_int(&mut self, vec_no: u8, _dpl: u8, _level: IntrLevel, handler: IntrHandlerFunc, name: &'static str)

/// Returns true during processing of an external interrupt and false at all other times.
pic.intr_context(&self) -> bool

/// During processing of an external interrupt, directs the interrupt handler to yield to a new
/// process just before returning from the interrupt. May not be called at any other time.
pic.intr_yield_on_return(&mut self)
```

## TODO:
* Testing
* Implement stubs
* Migrate timer interrupt 
* Integrate file system interrupts


kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+
+        let e0 = (function as usize) as u32 & 0xffff  // offset 15:0
+            | (SEL_KCSEG as u32) << 16; // Target code segment
+        let e1 = (function as usize) as u32 & 0xffff0000 // offset 31:16
+            | (1 << 15) // present
+            | (dpl as u32) << 13 // Descriptor privilege level
+            | (1 << 12) // System
+            | (gate_type as u32) << 8; // type
+
+        (e0 as u64) | ((e1 as u64) << 32)
```
1whatleytay, line 329: Maybe use our bitfield framework, `bitbybit` unless you have a great reason to do the bitops yourself?

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+        (pic.idt.len() * core::mem::size_of::<u64>() - 1) as u16,
+        pic.idt.as_ptr(),
+    );
+    asm!("lidt [{0}]", in(reg) &idtr_operand);
+
+    // Initialize interrupt names.
+    for i in 0..INTR_CNT {
+        pic.intr_names[i] = "unknown";
+    }
+    pic.intr_names[0] = "#DE Divide Error";
```
1whatleytay, line 148: Can this be a `match` in some `get_interrupt_name` function? Or make it a constant global (if rust likes it)?

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+/// Slave PIC - Command
+const PIC1_CTRL: u16 = 0xA0;
+/// Slave PIC - Data
+const PIC1_DATA: u16 = 0xA1;
+
+/// Number of x86 interrupt vectors.
+const INTR_CNT: usize = 256;
+
+/// Whether the interrupt is enabled.
+pub enum IntrLevel {
```
1whatleytay, line 31: There's a chunk of work here to work around the fact this is an enum (e.g. `if a == b { IntrOn } else { IntrOff }`). Is there another state you want to add in the future? Can't this be a `enabled: bool`?

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+    vec_no: u32,
+
+    /// Error code
+    error_code: u32,
+
+    /// Saved frame pointer (EBP)
+    frame_pointer: *mut core::ffi::c_void,
+
+    /// Next instruction to execute
+    eip: extern "C" fn() -> (),
```
1whatleytay, line 85: Making this a `fn` creeps me out a bit. It might point to code, but I can't actually invoke this normally without things catching on fire, right?

kernel/src/threading/interrupt.rs
ff-zhang: I think some sort of constant global var would be best.

kernel/src/threading/interrupt.rs
ff-zhang, line 1: Since this is going to be replacing the current implementation of the PIC and IDT, please also remove that portion of the current code base and update any dependencies to use your implementation.

kernel/src/threading/interrupt.rs
LanceLi1416: The primary reason is that Rust discourages modifying global static variables, and generally discourages the use of global variables.

As noted in line 108, the `intr_names` variable is used solely for debugging purposes and will be removed in future iterations. No core functionality or logic should interact with `intr_names`.

kernel/src/threading/interrupt.rs
LanceLi1416: Will do that in the next push.

kernel/src/threading/interrupt.rs
LanceLi1416: I agree, this approach is certainly suboptimal. I'm currently experimenting with the best method for passing functions and their addresses in Rust. Since, for now, only the file system hardware interrupt (which I'm responsible for) will interact with the PIC, I plan to determine the best practice during the integration of the file system interrupt.

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+    esp: *mut core::ffi::c_void,
+    /// Data segment for esp
+    ss: u16,
+}
+
+pub type IntrHandlerFunc = fn(frame: &IntrFrame);
+
+/// Programmable Interrupt Controller (PIC)
+pub struct Pic {
+    /// The Interrupt Descriptor Table (IDT).  The format is fixed by the CPU.  See [IA32-v3a]
```
ff-zhang, line 100: Where is `[IA32-v3a]`? I don't see a definition/link to it anywhere

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+        let handler = self.handlers[frame.vec_no as usize];
+        if let Some(handler) = handler {
+            // Handler is registered, invoke it
+            handler(frame);
+        } else if frame.vec_no == 0x27 || frame.vec_no == 0x2f {
+            // From PintOS:
+            // There is no handler, but this interrupt can trigger
+            // spuriously due to a hardware fault or hardware race
+            // condition.  Ignore it. */
+            // TODO: verify if this is true
```
ff-zhang, line 362: I don't believe this is possible as everything is emulated through Qemu.

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+            handlers: [None; INTR_CNT],
+            intr_names: ["unknown"; INTR_CNT],
+            unexpected_cnt: [0; INTR_CNT],
+            in_external_intr: false,
+            yield_on_return: false,
+        }
+    }
+}
+
+/// Initializes the PICs.  Refer to [8259A] for details.
```
ff-zhang, line 407: Same thing as before, please be explicit as to what `[8258A]` is.

kernel/src/threading/interrupt.rs
LanceLi1416: I was considering the possibility of supporting interrupt priority levels to prioritize the more crtitical interrupts. However, since we're not building a full-scale operating system, a simpler boolean flag would be a good idea. I'll make this adjustment in the next update.

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,455 @@
...
+///
+/// By default, interrupts 0...15 delivered by the PICs will go to interrupt vectors 0...15.
+/// Those vectors are also used for CPU traps and exceptions, so we reprogram the PICs so that
+/// interrupts 0...15 are delivered to interrupt vectors 32...47 (0x20...0x2f) instead.
+fn pic_init() -> Pic {
```
ff-zhang, line 412: I'm not sure if this the correct way to remap PICs. The guide you referenced (`https://wiki.osdev.org/8259_PIC`) describes it differently. In particular, I'm worried about the lack of `io_wait` calls.

kernel/src/threading/interrupt.rs
LanceLi1416: [Intel 8259A Programmable Interrupt Controller (8259A/8259A-2)](https://www.scs.stanford.edu/24wi-cs212/pintos/specs/8259A.pdf)

I will update these references in the next update/

kernel/src/threading/interrupt.rs
1whatleytay: You're right, but why would we want to mutate the interrupt names? I feel like those are baked into the processor really. Global variables are fine if they are immutable, no side effects to worry about.

kernel/src/threading/interrupt.rs
1whatleytay: Oh that's awesome! I feel like priority levels are a bit separate from an enabled property, but I'm fine with anything.

kernel/src/threading/interrupt.rs
LanceLi1416: I will remove this field from PIC.

kernel/src/threading/interrupt.rs
1whatleytay: I gotcha. I think void pointer or `usize` is fine for now, if you're okay with this. Rust is pretty picky with values that are technically invalid for their types (although the compiler probably isn't aggressive enough for that to matter here).

Plus when I see a `fn() -> ()` it makes me just want to... call it... but I know that's bad for me.

kernel/src/threading/interrupt.rs
LanceLi1416: Will do!

kernel/src/threading/interrupt.rs
LanceLi1416: I will double check.

kernel/src/threading/interrupt.rs
ff-zhang: For now, maybe changing the type to `fn() -> !` might be helpful

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,461 @@
...
+    ) {
+        if level {
+            unsafe {
+                self.idt[vec_no as usize] = self.make_trap_gate(INTR_STUBS[vec_no as usize], dpl);
+            }
+        } else {
+            unsafe {
+                self.idt[vec_no as usize] = self.make_intr_gate(INTR_STUBS[vec_no as usize], dpl);
+            }
+        }
```
1whatleytay, line 235: Nit: Maybe rename `level` to `enabled` if we are going to make it a `bool`? Optionally, maybe get rid of some of this repetition if you feel comfortable with this. 

```suggestion
        let stub = INTR_STUBS[vec_no as usize];
        
        unsafe {
            self.idt[vec_no as usize] = if level {
                self.make_trap_gate(stub, dpl);
            } else {
                self.make_intr_gate(stub, dpl);
            };
        }
```

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        self.handlers[vec_no as usize] = Some(handler);
+        self.intr_names[vec_no as usize] = name;
+    }
+
+    /// Registers external interrupt VEC_NO to invoke HANDLER, which is named NAME for debugging
+    /// purposes. The handler will execute with interrupts disabled.
+    pub fn intr_register_ext(&mut self, vec_no: u8, handler: IntrHandlerFunc, name: &'static str) {
```
1whatleytay, line 243: If all functions in `PIC` start with `intr_` maybe drop it? Use the extra characters to expand on some of your other abbreviations, like `register_external` or `register_internal`.

kernel/src/threading/interrupt.rs
```diff
@@ -0,0 +1,461 @@
...
+            | (dpl as u32) << 13 // Descriptor privilege level
+            | (1 << 12) // System
+            | (gate_type as u32) << 8; // type
+
+        (e0 as u64) | ((e1 as u64) << 32)
+    }
+
+    /// Creates an interrupt gate that invokes `function` with the given `dpl`.
+    fn make_intr_gate(&self, function: fn() -> (), dpl: u8) -> u64 {
+        self.make_gate(function, dpl, 14)
```
1whatleytay, line 340: Might be nice to make some of these magic numbers named constants. Is `14`/`15` a bit field?

kernel/src/threading/interrupt.rs
LanceLi1416: Good point. Will change the naming in a later update.

kernel/src/threading/interrupt.rs
LanceLi1416: Yes sorry I forgot to rename `level` to `enabled`. Will change that in a later update.

kernel/src/threading/interrupt.rs
LanceLi1416: Will make these magic numbers some constants.

ff-zhang:
Interrupt refactoring moved into https://github.com/KidneyOS/KidneyOS/pull/49.

PR #48, opened 2024-09-20
File system block layer

LanceLi1416:
The block layer for the file system. 

## API

```rust
/// Register a block device with the given `name`. The block device's `size` in sectors and its
/// `device_type` must be prvided, as well as the `driver` to access the block.
///
/// Returns the index of the block device.
BlockManager::register_block(
    &mut self,
    block_type: BlockType,
    block_name: &str,
    block_size: BlockSector,
    driver: BlockDriver,
) -> usize
```

```rust
/// Reads sector `sector` from the block device into `buf`, which must have room for
Block::read(&mut self, sector: BlockSector, buf: &mut [u8])

/// Writes sector `sector` from `buf`, which must contain `BLOCK_SECTOR_SIZE` bytes. Returns
/// after the block device has acknowledged receiving the data.
Block::write(&mut self, sector: BlockSector, buf: &[u8])

Block::get_index(&self)

Block::get_name(&self)

Block::get_size(&self)

Block::get_type(&self)
```

### DIFF
--- a/kernel/src/block/block_core.rs
+++ b/kernel/src/block/block_core.rs
@@ -1,8 +1,10 @@
 #![allow(dead_code)] // Suppress unused warnings
 
 use crate::drivers::dummy_device::DummyDevice;
+use alloc::boxed::Box;
 use alloc::{string::String, vec::Vec};
 use core::fmt;
+use core::result::Result;
 use kidneyos_shared::println;
 
 /// Size of a block device in bytes.
@@ -15,32 +17,35 @@ pub const BLOCK_SECTOR_SIZE: usize = 512;
 /// Good enough for devices up to 2 TB.
 pub type BlockSector = u32;
 
+/// Error type for block operations
+pub struct BlockError;
+
 /// Types of blocks
 #[derive(PartialEq, Copy, Clone)]
 pub enum BlockType {
     /// OS Kernel
-    Kernel(BlockSector),
+    Kernel,
     /// File system
-    FileSystem(BlockSector),
+    FileSystem,
     /// Scratch
-    Scratch(BlockSector),
+    Scratch,
     /// Swap
-    Swap(BlockSector),
+    Swap,
     /// "Raw" device with unidentified contents
-    Raw(BlockSector),
+    Raw,
     /// Owned by non-KidneyOS operating system
-    Foreign(BlockSector),
+    Foreign,
 }
 
 impl fmt::Display for BlockType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            BlockType::Kernel(_) => write!(f, "Kernel"),
-            BlockType::FileSystem(_) => write!(f, "File System"),
-            BlockType::Scratch(_) => write!(f, "Scratch"),
-            BlockType::Swap(_) => write!(f, "Swap"),
-            BlockType::Raw(_) => write!(f, "Raw"),
-            BlockType::Foreign(_) => write!(f, "Foreign"),
+            BlockType::Kernel => write!(f, "Kernel"),
+            BlockType::FileSystem => write!(f, "File System"),
+            BlockType::Scratch => write!(f, "Scratch"),
+            BlockType::Swap => write!(f, "Swap"),
+            BlockType::Raw => write!(f, "Raw"),
+            BlockType::Foreign => write!(f, "Foreign"),
         }
     }
 }
@@ -48,9 +53,28 @@ impl fmt::Display for BlockType {
 /// Lower-level interface to block device drivers
 pub trait BlockOp {
     /// Read a block sector
-    unsafe fn read(&self, sector: BlockSector, buf: &mut [u8]);
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the buffer is the correct size (i.e., `BLOCK_SECTOR_SIZE` bytes).
+    ///
+    /// The caller must also ensure that the sector is within the bounds of the block device.
+    ///
+    /// This function is labelled unsafe because the underlying driver implementation requires
+    /// the use of assembly code to access the hardware.
+    unsafe fn read(&mut self, sector: BlockSector, buf: &mut [u8]) -> Result<(), BlockError>;
+
     /// Write a block sector
-    unsafe fn write(&self, sector: BlockSector, buf: &[u8]);
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the buffer is the correct size (i.e., `BLOCK_SECTOR_SIZE` bytes).
+    ///
+    /// The caller must also ensure that the sector is within the bounds of the block device.
+    ///
+    /// This function is labelled unsafe because the underlying driver implementation requires
+    /// the use of assembly code to access the hardware.
+    unsafe fn write(&mut self, sector: BlockSector, buf: &[u8]) -> Result<(), BlockError>;
 }
 
 /// Supported block drivers
@@ -60,34 +84,9 @@ pub enum BlockDriver {
     Dummy(DummyDevice),
 }
 
-impl BlockDriver {
-    /// Unwrap the block driver to get the underlying block operation
-    fn unwrap(&self) -> &dyn BlockOp {
-        match self {
-            // TODO: Add drivers here
-            BlockDriver::Dummy(driver) => driver,
-        }
-    }
-
-    /// Read a block sector
-    fn read(&self, sector: BlockSector, buf: &mut [u8]) {
-        unsafe {
-            self.unwrap().read(sector, buf);
-        }
-    }
-
-    /// Write a block sector
-    fn write(&self, sector: BlockSector, buf: &[u8]) {
-        unsafe {
-            self.unwrap().write(sector, buf);
-        }
-    }
-}
-
 /// A block device
 ///
 /// **Note:** Once blocks are made they are immutable
-#[derive(PartialEq, Clone)]
 pub struct Block {
     /// Unique and immutable index of the block
     index: usize,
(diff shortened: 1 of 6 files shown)


kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// File system
+    FileSystem(BlockSector),
+    /// Scratch
+    Scratch(BlockSector),
+    /// Swap
+    Swap(BlockSector),
+    /// "Raw" device with unidentified contents
+    Raw(BlockSector),
+    /// Owned by non-KidneyOS operating system
+    Foreign(BlockSector),
```
1whatleytay, line 32: `BlockSector` is owned by each of these types. Probably best to move this out into a struct like:

```
struct TypedBlockOrSomething {
  type: BlockType,
  section: BlockSector
}
```

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+/// Supported block drivers
+#[derive(Clone, Copy, PartialEq)]
+pub enum BlockDriver {
+    // TODO: Add drivers here
+    Dummy(DummyDevice),
+}
+
+impl BlockDriver {
+    /// Unwrap the block driver to get the underlying block operation
+    fn unwrap(&self) -> &dyn BlockOp {
```
1whatleytay, line 65: !!

Some weirdness is happening here. You might need to explain some of this to me. Just going to list what I see:

 - `unwrap` typically means something like "get it or panic!" in Rust. But this doesn't panic. I think it's important to be consistent with how we use `unwrap` because right now those words set off alarm bells! Maybe something like `inner()`/`get_inner()`/`get_op()` would fit better or something.
 - Why are we doing this `&dyn BlockOp` + `enum` dance 😅? If we're going to use `dyn BlockOp`, why don't we just drop the enum entirely and use `Box<dyn BlockOp>` to own it (we have an allocator at this level, right?). If we're going to use an `enum`, then we should statically dispatch calls to `read` and `write` (maybe we don't need the trait at all).

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Verifies that `sector` is a valid offset within the block device.
+    ///
+    /// Panics if the sector is out of bounds.
+    fn check_sector(&self, sector: BlockSector) {
+        if sector >= self.block_size {
+            panic!(
+                "{}: Invalid sector {} (block size: {})",
+                self.block_name, sector, self.block_size
+            );
```
1whatleytay, line 129: Nit: it seems strange to call a method that's intended to panic if something goes wrong. There aren't great tools for recovering from a panic. It's probably better to return a `Result` and if you really want to panic then call `.unwrap()`.

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Get the block device with the given `index`.
+    ///
+    /// If the index is out of bounds, returns `None`.
+    pub fn by_id(&mut self, idx: usize) -> Option<&mut Block> {
+        if idx >= self.all_blocks.len() {
+            return None;
+        }
+
+        Some(&mut self.all_blocks[idx])
```
1whatleytay, line 250: `self.all_blocks.get(idx)`? Returns `None` if the index is out of bounds.

pommicket (commented):
good work overall!

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// Tha name of the block device
+    block_name: String,
+
+    /// The type of block
+    block_type: BlockType,
+    /// The block driver
+    driver: BlockDriver,
+
+    /// The size of the block device in sectors
+    block_size: BlockSector,
```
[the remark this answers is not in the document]
1whatleytay: Ah, it does. Gotcha.

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+}
+
+/// Lower-level interface to block device drivers
+pub trait BlockOp {
+    /// Read a block sector
+    unsafe fn read(&self, sector: BlockSector, buf: &mut [u8]);
+    /// Write a block sector
+    unsafe fn write(&self, sector: BlockSector, buf: &[u8]);
```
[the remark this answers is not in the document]
pommicket: Also, might be slightly nice to use `&mut self` here, just in case the device needs to change any of its state when these are called (`Block::read/write` takes `&mut self` so it wouldn't make a difference to the outer API).

kernel/src/block/block_core.rs
LanceLi1416: I didn’t use `Box` because I couldn’t find a way to import `std` :sob: . Apologies if I’m missing something obvious—any guidance would be really appreciated. Thanks!

kernel/src/block/block_core.rs
LanceLi1416: Oh this is a residual from a previous iteratioin. I will remove this in the upcoming update.

kernel/src/block/block_core.rs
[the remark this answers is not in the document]
LanceLi1416: Should be clearer after resolving https://github.com/KidneyOS/KidneyOS/pull/48#discussion_r1768854184. That said, I do think the name is a bit misleading—if you have any thoughts please let me know!

kernel/src/block/block_core.rs
pommicket: You should be able to get it from `alloc::boxed::Box` !

LanceLi1416: This worked, thank you!

kernel/src/block/block_core.rs
LanceLi1416: What about the current iteration?

kernel/src/block/block_core.rs
LanceLi1416: Fixed!

kernel/src/block/block_core.rs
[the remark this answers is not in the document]
LanceLi1416: Fixed!

kernel/src/block/block_core.rs
```diff
@@ -15,42 +17,64 @@ pub const BLOCK_SECTOR_SIZE: usize = 512;
...
-    unsafe fn read(&self, sector: BlockSector, buf: &mut [u8]);
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the buffer is the correct size (i.e., `BLOCK_SECTOR_SIZE` bytes).
+    ///
+    /// The caller must also ensure that the sector is within the bounds of the block device.
+    ///
+    /// This function is labelled unsafe because the underlying driver implementation requires
+    /// the use of assembly code to access the hardware.
```
1whatleytay, line 64: Cool, but I'm not sure if this warrants the `unsafe` keyword. Can't the driver check to make sure the buffer is the correct size and throw an error? Same for the bounds.

If the driver uses assembly, that doesn't mean that whole method needs to be unsafe. The `unsafe` keyword on a function usually means "there are certain conditions that need to be met for this function to work properly, and if you haven't checked them you will get `UB`".

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// Returns `true` if the sector is valid, `false` otherwise.
+    fn is_sector_valid(&self, sector: BlockSector) -> bool {
+        sector < self.block_size
+    }
+
+    /// Reads sector `sector` from the block device into `buf`, which must have room for
+    /// `BLOCK_SECTOR_SIZE` bytes.
+    pub unsafe fn read(&mut self, sector: BlockSector, buf: &mut [u8]) -> Result<(), BlockError> {
+        if !self.is_sector_valid(sector) || !Self::is_buffer_valid(buf) {
+            return Err(BlockError);
```
1whatleytay, line 129: I like this! I was going to suggest `read`/`write` to return an error anyway, because I figure users are going to have some control over the inputs to these methods (maybe).

Optional: You could make `BlockError` an enum and add cases for SectorValid/BlockError or cases for common driver issues outside of driver stuff. You might also want to `impl Error for BlockError { }`.

1whatleytay:
looking good :D

kernel/src/block/block_core.rs
LanceLi1416: Fixed!

kernel/src/block/block_core.rs
LanceLi1416: Added!

1whatleytay (approved):
👍 You should wait for someone else to sign off before merging though.

LanceLi1416:
> 👍 You should wait for someone else to sign off before merging though.

Awaiting endorcement from @pommicket

kernel/src/block/block_core.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// Read a block sector
+    fn read(&mut self, sector: BlockSector, buf: &mut [u8]) -> Result<(), BlockError>;
+
+    /// Write a block sector
+    fn write(&mut self, sector: BlockSector, buf: &[u8]) -> Result<(), BlockError>;
+}
+
+/// Supported block drivers
+#[derive(Clone, Copy, PartialEq)]
+pub enum BlockDriver {
```
pommicket, line 62: We don't need this anymore right?

pommicket: Other than that, looks good!

pommicket (commented):
Looks good!

kernel/src/block/block_core.rs
LanceLi1416: Apologise for the oversight. It's now removed.

PR #49, opened 2024-09-20
Refactor interrupt-related code

ff-zhang:
Currently, the code which setups and handles our interrupt flow is scattered around various files which aren't all named accordingly. This just refactors all the interrupt-related code into one module. This should also simplify interacting with it for people unfamiliar with the process system as there is now a clear place to setup interrupt handlers.

### DIFF
--- a/kernel/src/interrupts/timer.rs
+++ b/kernel/src/interrupts/timer.rs
@@ -1,4 +1,5 @@
-use crate::{sync::irq::MutexIrq, threading::scheduling::scheduler_yield_and_continue};
+use super::mutex_irq::MutexIrq;
+use crate::threading::scheduling::scheduler_yield_and_continue;
 use core::{arch::asm, time::Duration};
 use kidneyos_shared::serial::{inb, outb};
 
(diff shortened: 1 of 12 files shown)


kernel/src/interrupts/timer.rs
```diff
@@ -0,0 +1,36 @@
...
+
+static SYS_CLOCK: MutexIrq<Duration> = MutexIrq::new(Duration::new(0, 0));
+
+pub fn step_sys_clock() {
+    let mut clock = SYS_CLOCK.lock();
+    match clock.checked_add(TIMER_INTERRUPT_INTERVAL) {
+        Some(update) => {
+            *clock = update;
+        }
+        None => panic!("System clock overflowed!"),
```
1whatleytay, line 19: Yeah, something has probably gone horribly wrong if the system clock overflowed.

1whatleytay (approved):
LGTM, I understand this is tagged as a draft :eyes:

ff-zhang:
Yeah, I still want to see if I can abstract away a few of the interrupt implementation details from outside functions.

PR #50, opened 2024-09-23
ELF parsing with nom

1whatleytay:
I had some problems with our old ELF parser, and I'm working to make some sections of the thread starting code a bit better. Generally, this PR intends to do a few things with elf parsing:

 1. Don't use `unsafe` operations to interpret ELF files. If we're using Rust for our language of choice, we should leverage its safety to prevent memory safety issues.
 2. Be more robust with our validation. For example, I think the old parser would read uninitialized memory with an empty elf file.
 3. Use a parsing framework to make parsing binaries more consistent (e.g. on hypothetical platforms that have different padding rules).
 
Some notes: ELF parsing will aggressively fail when it sees something it doesn't recognize (e.g. a new program segment type). I think this means an ELF parsing error should not be taken as a sign that the ELF file is invalid, but instead that the ELF parser might need to be extended.


kernel/src/user_program/virtual_memory_area.rs
```diff
@@ -1,61 +0,0 @@
...
-    pub vm_end: usize,   // VMA end, exclusive
-    pub offset: usize,
-    pub flags: VmFlags,
-    // TODO: Add other necessary fields here
-}
-
-/* Stores boolean flags. The flags include read, write, execute, shared, and private, each represented as a single bit.
-This allows the operating system to track the properties and permissions of different memory areas allocated to a process.*/
-#[derive(Debug, PartialEq, Eq, Default)]
-pub struct VmFlags {
```
1whatleytay, line 42: Dropped, because this file and it's associated trait seemed to be exclusively used for representing ELF files.

kernel/src/user_program/elf.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub program_type: ElfProgramType,
+
+    // Offset of the program data within the file.
+    pub file_offset: u32,
+
+    // Mount address, virtual address.
+    pub virtual_address: u32,
+
+    // Typically ignored, unless our operating system cares about physical addresses.
+    // I fing this is typically equal to the virtual_address field.
```
ff-zhang, line 182: typo

kernel/src/user_program/elf.rs
```diff
@@ -0,0 +1,296 @@
...
+    pub section_header_count: u16,
+    pub section_header_index: u16,
+}
+
+impl ElfHeader {
+    pub fn parse(bytes: &[u8]) -> IResult<&[u8], ElfHeader> {
+        let (bytes, _) = tag([0x7F, b'E', b'L', b'F'])(bytes)?;
+
+        // Elf Bit Width, we don't parse 64-bits ELF binaries.
+        let (bytes, _) = tag([1])(bytes)?;
```
ff-zhang, line 83: Is there a way to guard against 64-bit ELF bins?

ff-zhang (commented):
LGTM! nom is a sweet parser

kernel/src/user_program/elf.rs
1whatleytay: Yep! `tag` will fail if the file doesn't match its input, so this line will throw if we get a 64-bit elf (1 -> 32bit, 2 -> 64bit).

kernel/src/user_program/elf.rs
1whatleytay: Fixed!

ff-zhang (approved):
Meant to leave an approving change, oops

kernel/src/user_program/elf.rs
1whatleytay: It's annoying because the 64-bit targeting file has a completely different layout 🥲 It's not just that the fields are larger, their order is swapped and everything.

PR #51, opened 2024-09-23
ATA/IDE device driver

LanceLi1416:
Driver for ide device.


kernel/src/block/block_error.rs
```diff
@@ -7,13 +7,19 @@ pub enum BlockError {
...
+    /// Error writing to the disk
+    WriteError,
 }
 
 impl Debug for BlockError {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
         match self {
             BlockError::SectorOutOfBounds => write!(f, "SectorOutOfBounds"),
             BlockError::BufferInvalid => write!(f, "BufferInvalid"),
+            BlockError::ReadError => write!(f, "ReadError"),
```
pommicket, line 21: BTW you can just add `#[derive(Debug)]` before `pub enum BlockError {` to implement this trait automatically

shared/src/serial.rs
```diff
@@ -29,6 +29,54 @@ pub unsafe fn inb(port: u16) -> u8 {
     res
 }
 
+/// Wrapper for assembly function insw - input from port to string.
+///
+/// Input word from I/O port specified in DX into memory location specified in ES:EDI.
+///
+/// # Safety
+///
+/// This function is unsafe because it uses inline assembly.
```
pommicket, line 38: Just using inline assembly is not a reason a function should be marked unsafe (you can always wrap the assembly in `unsafe { ... }` if it is safe). Better would be to say buffer[0..count] must be writable and port must be a valid I/O port or whatever.

kernel/src/drivers/ata/ata_timer.rs
```diff
@@ -0,0 +1,53 @@
...
+fn usleep_block(t: u64) {
+    for _ in 0..9 * t {
+        nsleep_block(t);
+    }
+}
+
+/// Sleep for `t` nanoseconds.
+fn nsleep_block(_t: u64) {
+    unsafe {
+        asm!("nop");
```
pommicket, line 24: Should there be something here that uses t?? I don't think we should rely on nop taking a specific amount of time. Maybe you can do `rdtsc` in a loop and determine when `t * K` cycles have passed, for K = cpu frequency in GHz (perhaps you can just pick K = 2 if precision isn't important here).

kernel/src/drivers/ata/ata_timer.rs
```diff
@@ -0,0 +1,53 @@
+// Wrapper for timer functions.
+
+use crate::interrupts::timer::sleep;
+use core::arch::asm;
+use core::time::Duration;
+
+/// Sleep for `t` milliseconds.
+fn msleep_block(t: u64) {
+    for _ in 0..9 {
```
pommicket, line 9: a..b does not include b in Rust. So this should be 0..10, or rather 0..1000 since there are 1000 μs in a ms, not 10. That said, I think it would be better to have this function just call `usleep_block(1000 * t)` (and likewise have `usleep_block` just call `nsleep_block(1000 * t)` — assuming `nsleep_block` actually looks at its parameter!

kernel/src/drivers/ata/ata_channel.rs
```diff
@@ -0,0 +1,587 @@
...
+            d1_name,
+            d1_is_ata: false,
+            channel_num,
+        }
+    }
+
+    /// Sets the name, d0_name, and d1_name to the appropriate values based on the channel number.
+    pub fn set_names(&mut self) {
+        // ascii '0' is 0x30
+        let name_char = char::from(0x30 + self.channel_num);
```
pommicket, line 515: btw you can use `b'0'` here to get the ASCII code for 0 as a u8. Then you don't need the comment before it!

pommicket (commented):
Looks good! Just a few issues with the blocking sleep code (maybe what you have there is just a placeholder though?)

Makefile
```diff
@@ -91,7 +91,10 @@ $(ISO): build/isofiles/boot/kernel.bin build/isofiles/boot/grub/grub.cfg
 run-bochs: $(ISO)
 	bochs -q -f bochsrc.txt
 
-QEMU_FLAGS := -no-reboot -no-shutdown -m 4G -d int,mmu,pcall,cpu_reset,guest_errors -cdrom $(ISO)
+# QEMU_FLAGS := -no-reboot -no-shutdown -m 4G -d int,mmu,pcall,cpu_reset,guest_errors -cdrom $(ISO)
+QEMU_FLAGS := -no-reboot -no-shutdown -m 4G -d int,mmu,pcall,cpu_reset,guest_errors -cdrom $(ISO) \
+			  -drive format=raw,file=gpt_fat16_50mb.img,if=ide \
```
1whatleytay, line 96: :eyes: We're targeting FAT16?

kernel/src/drivers/ata/ata_timer.rs
1whatleytay: `0..=9`?

shared/src/serial.rs
LanceLi1416: Understood! Will change in the next update.

kernel/src/drivers/ata/ata_timer.rs
LanceLi1416: I'll update per https://github.com/KidneyOS/KidneyOS/pull/51#discussion_r1776102909 so this issue will go away.

kernel/src/drivers/ata/ata_timer.rs
LanceLi1416: Will change in the next update.

kernel/src/main.rs
```diff
@@ -68,12 +69,16 @@ extern "C" fn main(mem_upper: usize, video_memory_skip_lines: usize) -> ! {
...
         let block_manager = block_init();
-        println!("{}", block_manager);
         println!("Block layer set up!");
 
         println!("Initializing Thread System...");
         thread_system_initialization();
         println!("Finished Thread System initialization. Ready to start threading.");
+
+        println!("Setting up IDE");
+        let _block_manager = ide_init(block_manager, true);
```
1whatleytay, line 79:
```suggestion
        let block_manager = ide_init(block_manager, true);
```

I mean, `block_manager` is moved here anyway.

1whatleytay (approved):

Makefile
LanceLi1416: I believe that is the file system we have so far? I'll add a script to generate the disk image at #53.

kernel/src/main.rs
LanceLi1416: Oh this is for pacifying the linter so it does not block the commit.
```
   = note: `-D unused-variables` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_variables)]`
```

PR #57, opened 2024-09-27
Create process table

mckeenicholas:
Creates a basic data structure to keep ownership of PCBs


gibinit
1whatleytay, line 1: Probably don't want to commit these two.

1whatleytay (approved):

kernel/src/threading/scheduling/wait_queue.rs
ff-zhang, line 1: Probably showed be renamed, bit of a misnomer to call this a queue

programs/wait/.gitignore
ff-zhang, line 1: Should this be included?

programs/syscall/syscall
ff-zhang, line 1: Binaries shouldn't be included

programs/wait/wait.S
ff-zhang, line 1: If this is only for testing of wait, I don't think we should include it

ff-zhang (commented):
Outside of some extraneous files, it looks good

kernel/src/threading/process_table.rs
```diff
@@ -0,0 +1,35 @@
...
+        assert!(
+            !self.table.contains_key(&pcb.pid),
+            "PCB with pid {} already added to process table.",
+            pcb.pid
+        );
+        self.table.insert(pcb.pid, pcb);
+    }
+
+    pub fn remove(&mut self, pid: Pid) -> Option<Box<ProcessControlBlock>> {
+        #[allow(dead_code)]
```
ff-zhang, line 27: NIT: Should this be above the function declaration?

ff-zhang (approved):
LGTM

kernel/src/threading/process_table.rs
mckeenicholas: I was getting intellisense errors with it above, this is just to make clippy not complain and can be removed once the function is used.

PR #58, opened 2024-10-01
Feat/thread sleep

mckeenicholas:
Adds thread sleep and wakeup functions and a mutex using a sleeping lock.

### DIFF
--- a/kernel/src/sync/mutex/sleep.rs
+++ b/kernel/src/sync/mutex/sleep.rs
@@ -19,10 +19,16 @@ pub struct SleepMutexGuard<'a> {
 }
 
 impl<'a> SleepMutexGuard<'a> {
-    pub fn unlock(mut self) {
+    pub fn unlock(&mut self) {
+        intr_disable();
         if let Some(mutex) = self.mutex.take() {
             mutex.unlock();
         }
+        intr_enable();
+    }
+
+    pub fn is_locked(&self) -> bool {
+        self.mutex.is_some()
     }
 }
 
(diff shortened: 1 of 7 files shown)


kernel/src/sync/mutex/sleep.rs
ff-zhang, line 1: If we add a `mutex` folder, should we move `mutex.rs` into it?

kernel/src/threading/scheduling/scheduler.rs
```diff
@@ -10,4 +10,5 @@ pub trait Scheduler {
     fn push(&mut self, thread: Box<ThreadControlBlock>);
     fn pop(&mut self) -> Option<Box<ThreadControlBlock>>;
     fn remove(&mut self, tid: Tid) -> Option<Box<ThreadControlBlock>>;
+    fn get(&mut self, tid: Tid) -> Option<&mut ThreadControlBlock>;
```
ff-zhang, line 13: In Rust, I think this should be name `get_mut`

ff-zhang (approved):
LGTM but I didn't test `SleepMutex`; it would be a good idea to wait for someone to review before merging

kernel/src/sync/mutex/sleep.rs
```diff
@@ -0,0 +1,154 @@
...
+            thread_wakeup(next_thread);
+        }
+
+        self.aquired = false;
+        self.holding_thread = None;
+        intr_enable();
+    }
+
+    pub fn is_locked(&mut self) -> bool {
+        self.aquired
```
1whatleytay, line 116: `&mut self`? Is this just for exclusivity?

1whatleytay (approved):

PR #59, opened 2024-10-02
Cargo Defaults, Syscalls Stubs and C/Rust Binaries

1whatleytay:
This PR changes a few things about the project:

 - The default linker for `cargo build` is now `i686-unknown-linux-gnu-cc`. This means `cargo build` will now succeed by default. Hooray!
 - Added a syscall stubs crate to make it easier to build more complex binaries, called `kidneyos-syscalls`.
 - Added `cbindgen` to the syscall stubs crate, so the same stubs can be used in C programs.
 - Added an example Rust and C binary that use the syscalls crate.
 - Modified behaviour of ELF files to support some more Program Header types, and to allow for Relocatable Executables (which have the "Shared" ELF type).

⚠️ **Note**: The linker change might affect the WSL environment (unsure if it broke or not).


programs.mk
```diff
@@ -1,14 +1,23 @@
...
-	$(AS) -o $@ $<
+exit:
+	cd programs/exit && make
+
+example_c:
+	cd programs/example_c && make
+
+example_rust:
+	# We don't want to export CARGO_TARGET_DIR to our destination make.
+	unset CARGO_TARGET_DIR && cd programs/example_rust && make
```
1whatleytay, line 16: Looking for some feedback on these Makefiles. My main goal was to have the files for building the binary independent from the rest of the project (e.g. one makefile per binary). This might have messed with Makefile reasoning though.

kernel/src/threading/thread_control_block.rs
```diff
@@ -107,79 +139,39 @@ impl ProcessControlBlock {
...
-                    kernel_virt_addr.add(program_header.data.len()),
+                    kernel_virt_addr.add(segment_padded_size),
                     0,
-                    frames * PAGE_FRAME_SIZE - program_header.data.len(),
+                    frames * PAGE_FRAME_SIZE - segment_padded_size,
                 );
             }
         }
 
-        let mut new_pcb = Self {
```
mckeenicholas, line 133: Are we not creating a new PCB anymore? Or is there a plan to move this elsewhere?

1whatleytay: I mean this was just discarded right? We should create this when we know how and when we are going to use a PCB.

mckeenicholas (approved):
LGTM!

ff-zhang (commented):
LGTM

ff-zhang (approved):

PR #61, opened 2024-10-03
Read-only FAT VFS implementation

pommicket:
Add read-only VFS implementation for FAT-16 and FAT-32. This will let us read files from the host file system, since qemu offers an option to mount a local directory as a FAT-formatted IDE device.


pommicket:
(Will un-draft this once we find a way of safely accessing block devices e.g. via #64 )

1whatleytay (approved):
That's a huge directory!!

PR #63, opened 2024-10-03
Add some basic functions for accessing userspace pointers

pommicket:
Add some functions so that we can access userspace pointers passed in from syscalls. Currently this is incomplete, since we need to check whether the pointers are actually mapped (e.g. on Linux at least, `open((void *)0x1234, 0)` returns `-EFAULT` rather than causing an interrupt). But it's good to at least have a version we can use for now.

Also I'm assuming we're following the Linux convention of pointers `>= 0x8000_0000` corresponding to kernel-space, but I'm not sure if that's true (seems to be as far as I can tell).

### DIFF
--- a/kernel/src/mem/util.rs
+++ b/kernel/src/mem/util.rs
@@ -1,0 +1,135 @@
+use crate::paging::{is_userspace_readable, is_userspace_writeable};
+
+pub enum CStrError {
+    Fault,
+    BadUtf8,
+}
+
+/// Minimum possible page size. It's okay for this to be smaller than the actual page size.
+const MIN_PAGE_SIZE: usize = 4096;
+
+#[must_use]
+fn can_access_address_range(start: *const u8, bytes: usize, write: bool) -> bool {
+    let check = if write {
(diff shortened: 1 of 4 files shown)


kernel/src/mem/util.rs
```diff
@@ -0,0 +1,135 @@
+use crate::paging::{is_userspace_readable, is_userspace_writeable};
+
+pub enum CStrError {
+    Fault,
+    BadUtf8,
```
ff-zhang, line 5: NIT: should it be "UTF" here?

ff-zhang (approved):
What's the reason we're using `u8` for the pointers? Otherwise, this looks good to me

kernel/src/mem/util.rs
pommicket: Good point - those functions can also be generic. As for `BadUtf8`, I am just following [std::str::Utf8Error](https://doc.rust-lang.org/std/str/struct.Utf8Error.html) (at least in the rust stdlib, acronyms aren't capitalized)

PR #64, opened 2024-10-04
Make `Block::{read, write}` safe by adding interrupt level check

pommicket:
We have a `intr_get_level` function to check whether interrupts are enabled, so I think it's better to panic when `Block::read` / `Block::write` are called with interrupts disabled, instead of having them be unsafe — then we would need to mark all filesystem functions as unsafe (or else add the interrupt check in all of those functions) since they call the Block methods. But let me know if there's any problem with this.

### DIFF
--- a/kernel/src/block/block_core.rs
+++ b/kernel/src/block/block_core.rs
@@ -1,6 +1,7 @@
 #![allow(dead_code)] // Suppress unused warnings
 
 use crate::block::block_error::BlockError;
+use crate::interrupts::{intr_get_level, IntrLevel};
 use alloc::boxed::Box;
 use alloc::{string::String, vec::Vec};
 use core::fmt;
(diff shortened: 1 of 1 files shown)


LanceLi1416:
I completely agree with your point! Checking the interrupt status within the block layer would definitely be helpful.

Actually, which approach do you think would be better:
1. Panic when interrupts are disabled.
2. Explicitly enable interrupts in `read` and `write`:
    ```rust
    pub fn read(&mut self, sector: BlockSector, buf: &mut [u8]) -> Result<(), BlockError> {
            let intr_level = intr_get_level();
            if intr_level == IntrOff { intr_enable(); }
            
            ...

            let ret = self.driver.read(sector, buf);
            if intr_level == IntrOff { intr_disable(); }
            ret
    }
    ```

LanceLi1416 (approved):
LGTM!

pommicket:
> Actually, which approach do you think would be better:

I think panicking makes more sense since usually when we turn off interrupts it's for an important reason (e.g. critical section that must not be interrupted), and we don't want to accidentally introduce bugs because we expect interrupts to be off when they're not.

PR #65, opened 2024-10-04
Add PCB functionality back

ff-zhang:
Creates the PCB when necessary and ownership of them is handed to a global `PROCESS_TABLE`.

### DIFF
--- a/kernel/src/threading/thread_control_block.rs
+++ b/kernel/src/threading/thread_control_block.rs
@@ -1,5 +1,5 @@
 use super::thread_functions::{PrepareThreadContext, SwitchThreadsContext};
-use crate::threading::process_table::ProcessTable;
+use crate::threading::process_table::PROCESS_TABLE;
 use crate::user_program::elf::{ElfArchitecture, ElfProgramType, ElfUsage};
 use crate::{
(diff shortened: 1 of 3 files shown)


ff-zhang:
Let me know what you think is missing from the implementation

kernel/src/threading/thread_control_block.rs
```diff
@@ -69,6 +61,23 @@ pub struct ProcessControlBlock {
...
+    pub fn create() -> Pid {
+        let pid = allocate_pid();
+        let pcb = Self {
+            pid,
+            child_tids: Vec::new(),
+            wait_list: Vec::new(),
+            exit_code: None,
+        };
+        unsafe {
+            PROCESS_TABLE.as_mut().expect("").add(Box::new(pcb));
```
1whatleytay, line 74: `.unwrap()`?

ff-zhang:
I believe the rest of the places where we would create a PCB aren't merged into `main` yet.

mckeenicholas (approved):

mckeenicholas:
Processess will have to be removed as well when the thread is reaped.

PR #67, opened 2024-10-07
Swap underlying bitfield crate 

ff-zhang:
We want to swap from the [bitbybit](https://crates.io/crates/bitbybit) crate over to a custom bitfield implementation.


ff-zhang:
@1whatleytay @pommicket The `bitfield!` macro is of the form:
```[rust]
bitfield! {
    struct Foo(u16);
    impl Debug;
    u16, bar, set_bar: 15, 0;
    ...
}
```
and generates a struct `Foo` with a getter `bar` and setter `set_bar`. For each setter, I want to write a function
```[rust]
fn with_bar(&mut self, value: T) {
    self.set_bar(value);
    self
}
```
so that we can chain multiple of them together. Do you know how to write a macro for this?

pommicket:
Hmm you can do
```rust
macro_rules! impl_with {
    ($type: ty, $setter: ident, $with: ident) => {
        fn $with(&mut self, value: $type) -> &mut Self {
            self.$setter(value);
            self
        }
    }
}

impl Foo {
    impl_with!(u16, set_bar, with_bar);
    // ... repeat for each field ...
}
```
But I think to do anything more automatic than this we'd need to modify the bitfield crate.

1whatleytay:
We can probably avoid chaining altogether with https://docs.rs/bitfield/latest/bitfield/macro.bitfield_constructor.html?

shared/src/bit_array.rs
```diff
@@ -0,0 +1,95 @@
...
+    };
+}
+
+impl_bitarray!({u8, u16, u32, u64} {u8, u16, u32, u64});
+
+#[macro_export]
+macro_rules! impl_getter_setter {
+    ($t: ty, $name: ident, $low: literal, $high: literal) => {
+        paste! {
+            pub fn $name(&self) -> $t { self.0.get_range($low, $high) as $t }
```
pommicket, line 37: These functions can be const right?

kernel/Cargo.toml
```diff
@@ -5,9 +5,8 @@ edition = "2021"
 
 [dependencies]
 arbitrary-int = "1.2.7"
```
pommicket, line 7: Looks like we can also remove `arbitrary-int` as a dependency now

shared/Cargo.toml
```diff
@@ -5,5 +5,5 @@ edition = "2021"
 
 [dependencies]
 arbitrary-int = "1.2.7"
```
pommicket, line 7: (We can also remove `arbitrary-int` here)

pommicket (approved):
LGTM!

shared/src/bit_array.rs
```diff
@@ -0,0 +1,95 @@
...
+                    (self.0 & mask) >> low as $v
+                }
+
+                pub const fn with(self, value: bool, idx: usize) -> Self {
+                    BitArray(self.0 | (value as $t) << idx)
+                }
+
+                pub const fn with_range(self, value: $v, low: usize, high: usize) -> Self {
+                    let mask = ((1 << high - low) - 1 | 1 << high - low) << low;
+                    BitArray((self.0 & !mask) | (value as $t & mask) << low)
```
pommicket, line 22: Actually this should probably be `(value << low) & mask` here right? (Also, we might want to have an assertion that `((value << low) & !mask) == 0`)

pommicket (changes requested):
(see above)

shared/src/bit_array.rs
ff-zhang: Nice catch

ff-zhang: Nice catch!

shared/src/bit_array.rs
ff-zhang: Those are no longer used and have been deleted.

pommicket (approved):
LGTM

PR #68, opened 2024-10-07
Add stub keyboard handler

pommicket:
Add an interrupt handler for keyboard interrupts, so that we don't panic when a key is pressed.


LanceLi1416 (approved):
LGTM

PR #70, opened 2024-10-07
WIP: Integrate SleepMutex into ATA/IDE Device Driver

LanceLi1416:
> **IMPORTANT:** A page table issue is currently blocking the creation of new threads, leading to the ide initialization being commented out in `main.rs`. Consequently, the IDE device driver is non-functional. Despite this, I've decided to merge the PR to avoid delaying the integration of other features.


kernel/src/drivers/ata/ata_channel.rs
```diff
@@ -571,15 +571,15 @@ impl AtaChannel {
         self.expecting_interrupt
     }
 
-    pub fn sem_down(&self) {
-        self.completion_wait.down();
+    pub fn sem_down(&mut self) {
+        self.completion_wait.lock();
```
mckeenicholas, line 575: The sleepmutex.lock() function returns a mutex guard, which will release the lock as soon as it's dropped. Since you're not saving it anywhere I don't think this will actually lock. 
The SleepMutex.unlock() method sohuld probably not be marked as public as it's only used internally.

pommicket: We should also probably mark the `SleepMutexGuard` with `#[must_use]`, to prevent errors like this. I think to have an interface like this that works properly you will need to add functions `raw_lock` and `raw_unlock` to `SleepMutex` which just directly lock/unlock the mutex without using a guard

PR #72, opened 2024-10-10
Add execve syscall implementation

1whatleytay:
This PR adds the execve syscalls. Since there isn't really a concept of processes, we just replace the current thread. Another note is we accept a ELF executable as bytes instead of a file path, since we don't have a filesystem set up.

The execve syscall does two things:

 - Parses an ELF file and creates a new thread.
 - Kills the current thread.

There's some memory read/write related calls, and I think @pommicket made a PR related to reading pointers. I haven't had the time to look at that, but I can port myself over to the new syscalls. Leave a comment!


ff-zhang (approved):
LGTM

PR #73, opened 2024-10-11
Add protected data structure to sleepmutex

mckeenicholas:
This adds a protected data structure that can only be accessed via the guard provided by the mutex.


pommicket (approved):
LGTM

PR #74, opened 2024-10-15
Integrate Sleep Mutex into IDE Driver

LanceLi1416:


kernel/src/drivers/ata/ata_core.rs
```diff
@@ -30,14 +28,9 @@ pub const ATA_IDENTIFY_DEVICE: u8 = 0xEC;
...
-    /// A list of ATA channels
-    pub static ref CHANNELS: Vec<Mutex<AtaChannel>> = {
-        (0..CHANNEL_CNT)
-            .map(|i| Mutex::new(AtaChannel::new(i as u8)))
-            .collect()
-    };
-}
+
+pub static mut CHANNELS: [AtaChannel; CHANNEL_CNT] = [AtaChannel::new(0), AtaChannel::new(1)];
+pub static mut ACCESS_MUTEX: [SleepMutex; CHANNEL_CNT] = [SleepMutex::new(), SleepMutex::new()];
```
pommicket, line 33: If #75 is approved, you should be able to change this to `pub static CHANNELS: [SleepMutex<AtaChannel>; CHANNEL_CNT] = [SleepMutex::new(AtaChannel::new(0)), SleepMutex::new(AtaChannel::new(1))]`, which will mean you can access them without unsafe code.
Otherwise LGTM.

LanceLi1416: Hi @pommicket @mckeenicholas,

I've encountered an issue with the new `SleepMutex` implementation (#75) where the mutex guard cannot be borrowed as mutable.

Here's an example:
```rust
pub static CHANNELS: [SleepMutex<AtaChannel>; CHANNEL_CNT] = [
    SleepMutex::new(AtaChannel::new(0)),
    SleepMutex::new(AtaChannel::new(1))
];

/// This will generate the error:
///
/// Cannot borrow immutable local variable `c` as mutable 
/// cannot borrow `*c` as mutable, as it is behind a `&` reference [E0596] 
/// `c` is a `&` reference, so the data it refers to cannot be borrowed as mutabl
fn do_something_1() {
    for (i, c) in CHANNELS.iter().enumerate() {
        let channel = &mut c.lock();
    }
}

/// This will generate the error
///
/// cannot borrow immutable static item `CHANNELS` as mutable [E0596] 
/// cannot borrow as mutable
fn do_something_2() {
    for (i, c) in CHANNELS.iter_mut().enumerate() {
        let channel = &mut c.lock();
    }
}
```

To solve the issue in `do_something_2`, I could declare `CHANNELS` as 
```rust
pub static mut CHANNELS: [SleepMutex<AtaChannel>; CHANNEL_CNT] = [
    SleepMutex::new(AtaChannel::new(0)),
    SleepMutex::new(AtaChannel::new(1))
];
```
However, this would require using an `unsafe` block, which we are trying to avoid. Wrapping `CHANNELS` in a Mutex would overcome this issue, but it locks all channels at once, which is **not the desired behaviour** -- we should only lock the individual channels.

Am I missing any alternatives here? If not, do we agree with the trade-off in `do_something_2`?

Thanks!

pommicket: `do_something_1` compiles without error for me… are you sure you have merged #75?
(If you have, you can also try
```rust
let mut guard = c.lock();
let channel = &mut guard;
```
)

LanceLi1416: Apologies, this seems to be a combined issue with `identify_ata_device`. I have pushed the problematic version to this branch so you can take a look. Really appreciate your help!

pommicket: It looks like this branch doesn't have the changes from #75 though. If I do `git merge main` on this branch, I only get the error that `force_unlock` is undefined (and also that `unlock` is private, but you shouldn't need the `channel.unlock();` at ata_core.rs:133 since the mutex will automatically be unlocked when the guard goes out of scope). Relatedly, I think we should try to avoid `force_unlock` if at all possible (which is why I didn't implement it for SleepMutex), because I don't think it can really be used in a safe way. But if it really is necessary we can implement it.

LanceLi1416: My apologies, I merged #73 but not #75. The `force_unlock` was a residual from the ticket mutex, which I will remove. Thank you so much for the help and truly sorry for the oversight.

pommicket (approved):
LGTM

PR #75, opened 2024-10-15
Make `SleepMutex::lock` take `&self`

pommicket:
Currently SleepMutex isn't very useful, since its lock method requires `&mut self`, and you can't have multiple mutable references at once to a value in Rust.
We can make `SleepMutex` take `&self` (as `std::sync::Mutex` does) by making the `holding_thread` atomic and using a `MutexIrq` to store the wait queue.
I have removed `force_unlock`, since `std::sync::Mutex` doesn't even have it (but lmk if you want to add it back in).

### DIFF
--- a/kernel/src/sync/mutex/sleep.rs
+++ b/kernel/src/sync/mutex/sleep.rs
@@ -1,12 +1,13 @@
 use crate::{
-    interrupts::{intr_disable, intr_enable},
+    interrupts::{intr_disable, intr_enable, mutex_irq::MutexIrq},
     threading::{
-        thread_control_block::Tid,
+        thread_control_block::{AtomicTid, Tid},
         thread_sleep::{thread_sleep, thread_wakeup},
         RUNNING_THREAD,
     },
 };
 use alloc::collections::VecDeque;
+use core::sync::atomic::Ordering::{AcqRel, Acquire, Release};
 use core::{
     cell::UnsafeCell,
     fmt,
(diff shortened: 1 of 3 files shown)


kernel/src/sync/mutex/sleep.rs
```diff
@@ -107,90 +106,80 @@ impl<T> SleepMutex<T> {
...
                 .expect("why is nothing running?")
                 .tid
         };
 
-        while self.is_locked() {
-            if !self.wait_queue.contains(&current_tid) {
-                self.wait_queue.push_back(current_tid);
+        loop {
+            // If no thread is holding the mutex, grab it.
+            let _ = self
```
mckeenicholas, line 120: Is this required since it's not marked as a #[must_use] and we're not checking if an error is returned?

pommicket: The function isn't marked `#[must_use]`, but the return type (`Result`) is. So we will still get a warning if we don't do this.

pommicket: ah okay it's because of `#[allow(unused)]` in the parent module (which lets us ignore `#[must_use]`). I have now replaced that with `#[allow(dead_code)]`.
Now if we remove the `let _ =` we do get a warning.

mckeenicholas:
LGTM!

mckeenicholas (approved):

PR #77, opened 2024-10-16
Implement NOOP Scheduler

LanceLi1416:
A NOOP scheduler that does not do any scheduling.


kernel/src/block/schedulers/noop_scheduler.rs
```diff
@@ -0,0 +1,39 @@
...
+        Self {
+            block,
+            queue: Vec::new(),
+        }
+    }
+}
+
+impl BioScheduler for NoopScheduler {
+    fn enqueue(&mut self, r: &mut BioReq) {
+        self.queue.push(r.clone());
```
pommicket, line 24: Probably either this or `wait` should be marked `unsafe` since we are not verifying that the BioReq is valid (e.g. buffer could be an invalid pointer or sector could be out-of-bounds)?

kernel/src/block/schedulers/noop_scheduler.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    fn wait(&mut self) -> Result<(), BlockError> {
+        let r = self.queue.pop().unwrap();
```
pommicket, line 28: Maybe it would be nice for this to return `Ok(false)` if there are no requests in the queue, rather than panicking (so we can do something like `while scheduler.wait()? {}` to wait for all requests to go through)

kernel/src/block/schedulers/noop_scheduler.rs
LanceLi1416: Update.

kernel/src/block/schedulers/noop_scheduler.rs
LanceLi1416: By design, reaching this point implies that there should always be at least one request in the queue -- the one we just added. If the queue is empty, it indicates a critical issue with the scheduler (fewer requests in the queue than threads waiting, which could cause some threads to wait indefinitely). In this case, I believe triggering a panic would be the most appropriate response. Let me know your thoughts!

pommicket: I see. But then is it ever possible for the queue to have multiple elements in it? If we always wait for something to be handled right after it is added, then I don't think so. Unless it's because multiple threads can make requests simultaneously (I'm assuming this is why we want scheduling in the first place but lmk if I'm wrong), but then we have the following issue: suppose we have
```
// Thread A
self.scheduler.enqueue(request 1)
// Context  switch to Thread B before Thread A gets to wait
self.scheduler.enqueue(request 2)
self.scheduler.wait()
// (request 1 finishes, so wait returns, but Thread B is now assuming that request 2 finished)
```

I think we will want to instead have `enqueue` return an ID, and pass that ID as a parameter to `wait`.

btw I think `BLOCK_MANAGER` should not be static mut. static muts are very hard to use safely — so much so that in the upcoming Rust 2024, [it will be an error to take a reference to a static mut!](https://doc.rust-lang.org/edition-guide/rust-2024/static-mut-reference.html) (in fact this code is currently unsound since multiple threads could call `read` on different block devices simulatenously, which could potentially cause multiple mut references to `BLOCK_MANAGER` to exist at once, which is illegal). We should instead use Mutexes to deal with shared state.
Here's what I'd recommend doing:
- put `driver` into a mutex and make `read_raw/write_raw` take `&self`
- put `scheduler.queue` into a mutex and make `enqueue`, `wait`, `read`, `write` take `&self`
- make `BlockManager::{by_id, by_name}` take `&self` and return `&Block` rather than `&mut Block`
- put `BLOCK_MANAGER` into a `RwLock` (adding new block devices is rare so this should be quite low-cost since multiple readers are allowed at once). Unfortunately we don't have a `RwLock` yet, so maybe you just have to use a Mutex for now.

LanceLi1416: I agree. In fact, I'm adding a request ID in the upcoming iterations, so `wait` will know which request it’s waiting for.

Regarding the `BLOCK_MANAGER`, I believe @1whatleytay  is moving it into a global system structure, which should be mutex-locked. I'll confirm with him during today's meeting. Thanks for your feedback!

PR #79, opened 2024-10-17
Feature: Semaphore

1whatleytay:
I changed the implementation of a semaphore to be inline with Tokio's semaphores (https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html).

I had some issues with our scheduler that I'm looking into now. This prevented me from testing Semaphores with multiple threads, so be a bit extra careful in your reviews. I assume problems might need to be diagnosed in IDE where it is used.

Example usage:

```rs
let semaphore = Semaphore::new(2);

// Take one. semaphore = 1
let p1 = semaphore.acquire();

// Take one. semaphore = 0
let p2 = semaphore.try_acquire().unwrap();

// Can't acquire any more, semaphore is empty.
let p3 = semaphore.try_acquire();
assert!(p3.is_none());

// Drop one permit.
drop(p1);

// We can grab one now.
let p4 = semaphore.try_acquire();
assert!(p3.is_some());
```

We can also `forget` semaphore permits, which means they will not increment the Semaphore count when dropped. And we can `post` the semaphore itself to increment the count manually.


1whatleytay:
Pretty sure this is a problem with the main branch, and not the semaphore structure. Unless @LanceLi1416 you have evidence against this.

LanceLi1416 (approved):
LGTM!

PR #80, opened 2024-10-17
Proposal: SystemState Struct

1whatleytay:
I want to avoid spreading state around the project, so I've moved some components of the project into a system state structure. This is more of a proposal to get some feedback on centralizing our state like this (any ideas on how we can do synchronization better? any components that should be managed by a SystemState struct? any way we can make working with this struct better?)

Why go with a single SystemState structure?
 - State isn't spread throughout the project. Resetting and reasoning about the system state can be centralized.
 - This encourages that we pass the system state properly by reference to methods that need to access it.
 - Reduces the amount of code that might need to deal with unsafe `static mut`.

The problem of synchronizing SystemState is pretty challenging, so I've kept it as `static mut` and left the responsibility of synchronization to the member `State` structs.

Components that were not moved:
 - IDT, GDT, or other objects that are passed or dealt with specifically by the CPU. I think the implications of "dropping" the IDT might be too much for me to deal with.


kernel/src/sync/mutex/sleep.rs
```diff
@@ -140,7 +138,11 @@ impl<T: ?Sized> SleepMutex<T> {
 
     fn unlock(&self) {
         let running_tid = unsafe {
-            RUNNING_THREAD
+            SYSTEM
```
mckeenicholas, line 141: Is there a reason we cant use unwrap_system() here?

1whatleytay: no

kernel/src/threading/mod.rs
```diff
@@ -1,69 +1,66 @@
...
+        // Acquiring mutable reference to system here...
+        // ... so we can discard it before we enable interrupts.
+        let system = unwrap_system_mut();
+
+        // We must 'turn the kernel thread into a thread'.
+        // This amounts to just making a TCB that will be in control of the kernel stack and will
+        // never exit.
+        // This thread also does not need to enter the `run_thread` function.
+        // SAFETY: The kernel thread's stack will be set up by the context switch following.
+        // SAFETY: The kernel thread is allocated a "fake" PCB with pid 0.
```
mckeenicholas, line 52: I believe with the new sleep mutex, the kernel thread was switched to 1 and 0 represents a null thread. Code should be fine, but the documentation may need to be updated.

1whatleytay: Oh! Why does `0` represent a null thread? What does the null thread mean?

ff-zhang: The PID `0` represents a kernel process, this means there isn't an actual PCB/process associated with it

ff-zhang (approved):
Process stuff LGTM. I'm not sure about FS code

mckeenicholas:
LGTM

PR #82, opened 2024-10-18
Revert "Proposal: SystemState Struct"

mckeenicholas:
Reverts KidneyOS/KidneyOS#80

Small error found in syscalls.rs and thread_control_block.rs causing code not to compile.

`ThreadControlBlock::new_from_elf()` takes 2 parameters, hoever the call to it in syscalls.rs doesnt provide the ProcessStruct

This PR doesn't have to be accepted, we just need to fix the one instance in syscall and re-merge the branch.


LanceLi1416:
See #83

PR #84, opened 2024-10-19
Add parent PID tracking, getpid/getppid

mckeenicholas:


kernel/src/threading/thread_control_block.rs
```diff
@@ -93,7 +96,7 @@ impl ThreadControlBlock {
             panic!("ELF was valid, but it was not an executable or it did not target the host platform (x86)");
         }
 
-        let pid: Pid = ProcessControlBlock::create(state);
+        let pid: Pid = ProcessControlBlock::create(state, 0);
```
LanceLi1416, line 99: It might be more flexible to implement a `get_ppid` function that currently returns `0`. This approach will make it easier to extend the functionality in the future when we add support for multiple threads.

LanceLi1416 (approved):
Otherwise LGTM

kernel/src/user_program/syscall.rs
```diff
@@ -67,9 +69,33 @@ pub extern "C" fn handler(syscall_number: usize, arg0: usize, arg1: usize, arg2:
 
             scheduler_yield_and_die();
         }
+        SYS_GETPID => {
+            let tcb = unsafe {
+                unwrap_system()
```
pommicket, line 74: It might be nice to split out the syscall implementations into functions `fn getpid() -> Pid { ... }`, `fn getppid() -> Pid { ... }` so that other parts of the kernel which need to access the PID/PPID can do so easily

kernel/src/threading/thread_control_block.rs
mckeenicholas: I've modified it to use the same parent pid as the current process a new thread should re-use the same pid/ppid.
If there's nothing running (such as spawning the kernel thread), then it just uses 0.

PR #85, opened 2024-10-19
Fix Residual Characters when Scrolling in VideoMemoryWriter

LanceLi1416:
> **Note:** The linter suggested using an iterator, so I tried the following:
> ```rust
>                 for char in video_memory.iter_mut().take(end).skip(start) {
>                     *char = Character {
>                         ascii: b' ',
>                         attribute: self.attribute,
>                     };
>                 }
> ```
> However, this didn't work as expected. Could someone suggest a fix?

Fix the issue in the `VideoMemoryWriter` where residual characters remain on the screen after scrolling.

Screenshot 2024-10-19 at 6 38 03 PM


pommicket:
Yeah that lint is not very helpful. What you really want is `for c in &mut video_memory[start..end]` (just committed this)

pommicket (approved):
LGTM now

PR #91, opened 2024-10-24
Add more RootFileSystem tests

pommicket:
Add more tests for the filesystem-related syscalls

### DIFF
--- a/kernel/src/fs/fs_manager.rs
+++ b/kernel/src/fs/fs_manager.rs
@@ -1199,7 +1199,9 @@ pub static ROOT: Mutex<RootFileSystem> = Mutex::new(RootFileSystem::new());
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::user_program::syscall;
     use crate::vfs::tempfs::TempFS;
+    use std::ffi::CStr;
     fn test_pcb(root: &RootFileSystem) -> ProcessControlBlock {
         ProcessControlBlock {
             pid: 0,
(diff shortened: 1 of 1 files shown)


kernel/src/fs/fs_manager.rs
```diff
@@ -1265,4 +1283,139 @@ mod test {
...
+                dirents.as_mut_ptr().cast(),
+                dirents.len() * std::mem::size_of_val(&dirents[0]),
+            )
+        }
+        .unwrap();
+        let mut offset = 0;
+        let mut entries = vec![];
+        let dirents_ptr: *const u8 = dirents.as_ptr().cast();
+        while offset < n {
+            let dirent_ptr: *const Dirent = unsafe { dirents_ptr.add(offset).cast() };
```
1whatleytay, line 1374: We can't turn `*const Dirent` to a `&Dirent` and just use dot syntax to grab its properties? Am I misreading this code?

1whatleytay (approved):
No reason to hold this back!

I was going to try to change other syscalls to use `*const c_char` a bit, since that's a bit nicer in our cbindgen output, but the CStr API doesn't really expose a safe way to deal with `c_char`s in my research.

kernel/src/fs/fs_manager.rs
pommicket: The problem is that the name is stored directly after the rest of the properties and variable-size structs aren't really supported well in Rust currently.
( `*dirent_ptr` is maybe clearer than read() though so I'll switch to that).

PR #96, opened 2024-10-24
Feat/clock gettime

mckeenicholas:
Add syscall to get the systems current time.

Currently only realtime and monotonic clocks (which uses tsc) are supported.


pommicket (approved):
LGTM

PR #100, opened 2024-10-28
Implement RwLock

pommicket:
Add an RwLock structure, similar to `std::sync::RwLock`.
I've tested out this implementation with 10000 threads on my machine and it seems to work fine.

This PR also removes `force_unlock` since clippy was complaining about it and we're not using it anymore (std's mutex doesn't have it so I think it's fine if we don't— forcibly unlocking a mutex is almost always a bad idea...)


mckeenicholas (approved):
LGTM!

PR #101, opened 2024-10-28
Accessing SystemState safely

### LINKED ISSUE #87: Safe way of accessing SystemState
Currently SystemState is implemented using a `static mut`, so it's nearly impossible to use it in a truly safe way.

We discussed the possibility of using OnceLock (or maybe once_cell::sync::OnceCell since that doesn't need std) last Thursday —  i.e. make SystemState only accessible via a shared reference, and use individual mutexes on anything that needs mutation. This lets multiple threads access the system simultaneously (without unsafe code), while still providing mutual exclusion when necessary. 

I know @1whatleytay has been working on setting up the SystemState so far, but I'd also be happy to work on this.

pommicket:
Resolves #87 
With this PR, you can only access SystemState through a shared reference. The members of SystemState have to manage shared access individually.
We still have to use a `static mut` for the SystemState, but it's now private and can only be accessed through safe functions (an atomic variable is used to keep track of whether the systesm is initialized). I wanted to use the `once_cell` crate, but it looks like that needs either std or critical-section, and critical-section wasn't working …

Also I've moved the `RootFileSystem` object to SystemState, now that it can be accessed safely there 
~~(Currently a draft until #100 is merged)~~

### DIFF
--- a/kernel/src/main.rs
+++ b/kernel/src/main.rs
@@ -28,7 +28,8 @@ extern crate alloc;
 
 use crate::block::block_core::BlockManager;
 use crate::drivers::ata::ata_core::ide_init;
-use crate::system::{SystemState, SYSTEM};
+use crate::sync::rwlock::sleep::RwLock;
+use crate::system::SystemState;
 use crate::threading::process::create_process_state;
 use crate::threading::thread_control_block::ThreadControlBlock;
 use alloc::boxed::Box;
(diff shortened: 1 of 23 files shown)


kernel/src/main.rs
```diff
@@ -92,22 +93,22 @@ extern "C" fn main(mem_upper: usize, video_memory_skip_lines: usize) -> ! {
...
+        root.mount_root(TempFS::new())
+            .expect("Couldn't mount root FS");
+
+        crate::system::init_system(SystemState {
             threads,
             process,
-
-            block_manager,
+            block_manager: RwLock::new(block_manager),
+            root_filesystem: Mutex::new(root),
```
1whatleytay, line 108: :heart:

1whatleytay (approved):

PR #102, opened 2024-10-29
Fix/process thread split

mckeenicholas:
This implements/fixes some of the behaviour we talked about last week regarding process/thread exiting and waiting.


kernel/src/user_program/syscall.rs
```diff
@@ -27,7 +27,8 @@ pub extern "C" fn handler(syscall_number: usize, arg0: usize, arg1: usize, arg2:
     // Translate between syscall names and numbers: https://x86.syscall.sh/
     match syscall_number {
         SYS_EXIT => {
-            thread_functions::exit_thread(arg0 as i32);
+            // thread_functions::exit_thread(arg0 as i32);
+            process_functions::exit_process(arg0 as i32);
```
ff-zhang, line 31: Leftover comment should probably be removed

ff-zhang (approved):
LGTM

PR #103, opened 2024-10-30
Implement RNG/getrandom syscall

mckeenicholas:
This PR introduces functionality to generate cryptographically secure random sequences using the RDRAND instruction, leveraging the CPU’s built-in hardware random number generator.

To enable this feature in QEMU, the CPU architecture had to be explicitly set, as it uses hardware-specific instructions. However, all Intel and AMD CPUs since 2015 support it, so this should work on most x86-based systems.

This feature should be tested on non-x86 CPUs to confirm compatibility before merging however.

### DIFF
--- a/kernel/src/user_program/random.rs
+++ b/kernel/src/user_program/random.rs
@@ -14,12 +14,8 @@ fn generate_random_i32() -> Option<i32> {
             options(nostack, nomem),
         );
     }
-
-    if success == 1 {
-        Some(random_int)
-    } else {
-        None
-    }
+    
+    (success == 1).then_some(random_int)
 }
 
 /// Fills a buffer with random bytes from the CPU's RDRAND instruction.
--- a/kernel/src/user_program/random.rs
+++ b/kernel/src/user_program/random.rs
@@ -1,8 +1,7 @@
 use core::arch::asm;
 use core::ptr;
 
-/// Generates a random `i32` using the CPU's RDRAND instruction.
-/// Returns `Some(random_value)` on success or `None` if RDRAND fails.
+/// Generates a random int using the CPU's RDRAND instruction.
 fn generate_random_i32() -> Option<i32> {
     let mut random_int: i32;
     let success: u8;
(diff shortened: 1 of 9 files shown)


kernel/src/user_program/random.rs
```diff
@@ -0,0 +1,60 @@
...
+        Some(random_int)
+    } else {
+        None
+    }
+}
+
+/// Fills a buffer with random bytes from the CPU's RDRAND instruction.
+/// Returns the number of bytes written, or -1 if an error occurs.
+/// Currently no flags are implemented, if there is no random data available,
+/// an error code is returned.
```
pommicket, line 29: Is this true? It looks like right now it will return 0 if no random data is available (I think that's fine, but it would be good to correct the comment then).

kernel/src/user_program/random.rs
```diff
@@ -0,0 +1,60 @@
...
+pub fn getrandom(buffer: *mut u8, length: usize, _flags: usize) -> isize {
```
pommicket, line 30: I think it would be better to have this function take a `&mut [u8]`. (Right now this function should be marked `unsafe` since it's dereferencing a pointer without checking it. But if it takes `&mut [u8]` then it can be safe)

kernel/src/user_program/random.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut bytes_written: usize = 0;
+    let chunks = length / 4;
+    let remainder = length % 4;
+
+    for i in 0..chunks {
+        match generate_random_i32() {
+            Some(random_int) => {
+                let random_bytes = random_int.to_le_bytes();
+                for (j, &byte) in random_bytes.iter().enumerate() {
```
pommicket, line 39: This can be simplified to `buffer.add(i * 4).copy_from_nonoverlapping(random_bytes.as_ptr(), 4)`. or better yet
`buffer[i*4..i * 4 + 4].copy_from_slice(&random_bytes)` if buffer is changed to a slice

pommicket (changes requested):

pommicket (approved):
LGTM

PR #106, opened 2024-10-31
Move tid/pid references to Rc<RefCell<>>

mckeenicholas:


pommicket:
BTW I think you should use `Arc<Mutex<>>` rather than `Rc<RefCell<>>` since `Rc` and `RefCell` aren't thread-safe (they don't implement Send/Sync)

PR #112, opened 2024-11-03
Keyboard support

LanceLi1416:
Adds keyboard support. To get the input buffer, 
```rust
unwrap_system().input_buffer.lock().getc() -> Option<u8>
```


pommicket (approved):
It might be nice to eventually use a Mutex or atomics for the modifier keys just so we don't need to resort to unsafe code. But this looks good for now!

PR #114, opened 2024-11-07
Swapping Functions Using Disk Partition

michaelzixizhou:
Implemented the `SwapSpace` pub struct that has the `swap_in` and `swap_out` functions that take in a swap index and a frame number. 

Utilizes the swap partition initialized on the current disk.


kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,118 @@
...
+const SWAP_SECTORS: u32 = 8192;
+const SWAP_SIZE: BlockSector = SECTOR_SIZE * SWAP_SECTORS;
+const FRAMES: usize = SWAP_SIZE as usize / PAGE_FRAME_SIZE;
+const SWAP_INDEX: usize = 3;
+const SECTORS_IN_PAGE: u32 = PAGE_FRAME_SIZE as u32 / SECTOR_SIZE;
+
+// TODO: Convert VA usizes to to VirtAddr after merge.
+const SWAP_START: BlockSector = 10240;
+
+// TODO: Convert bitmap to an actual bitmap.
```
ff-zhang, line 19: Use the bitfield macro in `kidneyos_shared` here.

kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,118 @@
...
+pub struct SwapSpace {
+    bitmap: Box<[u8; FRAMES]>,
+    mem_start: *mut u8,
+    mem_limit: *mut u8,
+}
+
+// Bit 8 for swap bit.
```
ff-zhang, line 26: This should be 9

kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,118 @@
...
+
+// TODO: Convert VA usizes to to VirtAddr after merge.
+const SWAP_START: BlockSector = 10240;
+
+// TODO: Convert bitmap to an actual bitmap.
+pub struct SwapSpace {
+    bitmap: Box<[u8; FRAMES]>,
+    mem_start: *mut u8,
+    mem_limit: *mut u8,
+}
```
ff-zhang, line 24: Add a TODO to check if it is necessary to hold a pointer here (instead of a `u8` of the memory address)

ff-zhang (changes requested):
I looked over everything except the `swap_in` and `swap_out` functions.

kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,119 @@
...
+
+// Bit 9 for swap bit.
+impl SwapSpace {
+    /// # Safety
+    ///
+    ///  mem_start + SWAP_SIZE is within bounds.
+    pub unsafe fn new(mem_start: *mut u8) -> Self {
+        let mem_limit = unsafe { mem_start.add(SWAP_SIZE as usize) };
```
tonyhu-x, line 33: As we discussed, the calculation here may be wrong.

kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,119 @@
...
+            unsafe {
+                copy_nonoverlapping(buffer_ref.as_ptr(), frame_ptr, SECTOR_SIZE as usize);
+            }
+
+            // Increment pointer
+            frame_ptr = frame_ptr.add(SECTOR_SIZE as usize);
+        }
+
+        self.bitmap[swap_idx] = 0;
+    }
```
tonyhu-x, line 82: If we already have the frame, is it not possible to read directly into the frame?

kernel/src/swapping/swapping_utils.rs
```diff
@@ -0,0 +1,119 @@
...
+            swap.read(swap_sector + (i * SECTOR_SIZE), buffer_ref)
+                .expect("Block read error");
+
+            // Write sector to pointer
+            unsafe {
+                copy_nonoverlapping(buffer_ref.as_ptr(), frame_ptr, SECTOR_SIZE as usize);
+            }
+
+            // Increment pointer
+            frame_ptr = frame_ptr.add(SECTOR_SIZE as usize);
```
tonyhu-x, line 78: Here, as with below, it might be better to use pointer.byte_add to be clear that we're counting in units of bytes. It does the same thing here, since frame_ptr is a u8 pointer, but if that somehow changes, then this will cause a bug.

PR #115, opened 2024-11-07
Don't use indirect EIP for TCB

1whatleytay:
We have an interpretation problem for our `eip` variable in our TCB struct. There are two cases:

 - If pid == 0, this thread is considered a kernel thread, and `eip` is the pointer to be a **pointer to a pointer** to the entry of the thread.
 - If pid != 0, this thread is considered a user thread, and `eip` is considered to be a **pointer** to the entry of the thread.
 
 This PR decouples these behaviours and avoids some crashes with multiple TCBs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,11 @@ run-qemu-ng: $(ISO)
 	# NOTE: You can quit with Ctrl-A X
 	qemu-system-i386 -nographic $(QEMU_FLAGS)
 
+.PHONY: run-qemu-ng-gdb
+run-qemu-ng-gdb: $(ISO) build/kernel.sym
+	# NOTE: You can quit with Ctrl-A X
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -117,6 +117,11 @@ run-qemu-ng: $(ISO)
 	# NOTE: You can quit with Ctrl-A X
 	qemu-system-i386 -nographic $(QEMU_FLAGS)
 
+.PHONY: run-qemu-ng-gdb
+run-qemu-ng-gdb: $(ISO) build/kernel.sym
+	# NOTE: You can quit with Ctrl-A X
+	qemu-system-i386 -nographic -s -S $(QEMU_FLAGS)
```
1whatleytay, line 123: Handy make target I'd like to merge in.

LanceLi1416 (approved):
LGTM

PR #116, opened 2024-11-07
Read execve executable from disk

1whatleytay:
The `execve` syscall now uses the disk! Hooray!

`argv` and `envp` are ignored.


pommicket (approved):
Great!

PR #117, opened 2024-11-13
Add waitpid syscall

mckeenicholas:
Needs #113, 

Process groups and using a pid of 0 to wait on the child thread isn't implemented.

Tested by hardcoding multiple threads to be created on startup.


kernel/src/user_program/syscall.rs
```diff
@@ -54,7 +57,41 @@ pub extern "C" fn handler(syscall_number: usize, arg0: usize, arg1: usize, arg2:
...
+            if parent_pcb.waiting_thread.is_some() {
+                return -1;
+            }
+
+            parent_pcb.waiting_thread = Some(running_thread_tid());
+
+            // Disable interrupts as this must be done atomically
+            intr_disable();
+            if parent_pcb.exit_code.is_none() {
+                thread_sleep();
```
pommicket, line 85: Should there be a loop here? I don't see why exit_code would have to be set after thread_sleep()...
Also it's probably a good idea to drop the lock on parent_pcb before sleeping to avoid deadlocks.

pommicket (approved):
LGTM

PR #119, opened 2024-11-19
Add support for VMAs

pommicket:
Currently the only VMA is the stack, which is now allocated on-demand rather than all at once when the process is created.

The stack physical memory isn't freed on process exit yet, but it looks like that wasn't happening before, so at least it's not a regression.


mckeenicholas:
It might be worth adding some function to interate over the VMAs (such as for fork), but other than that, LGTM!

mckeenicholas (approved):

PR #126, opened 2024-11-27
Implement read-only VSFS

minjaedavidpark:
Refactoring && more tests to be done later during the winter break...


kernel/src/fs/vsfs/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+        Err(Error::ReadOnlyFS)
+    }
+
+    fn truncate(&mut self, _file: INodeNum, _size: u64) -> Result<()> {
+        Err(Error::ReadOnlyFS)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
pommicket, line 482: Maybe it's worth adding `#[allow(dead_code, unused_variables)]` here for now , to suppress the warnings for `make test`

kernel/src/fs/vsfs/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+
+            let inode_num = u32::from_le_bytes(slice[0..4].try_into().unwrap());
+            if inode_num < 0x8000 {
+                // First get the file name
+                let file_name = &slice[4..];
+                let name_index = names.len();
+
+                let mut file_name_str = String::new();
+                #[allow(clippy::needless_range_loop)]
+                for i in 0..file_name.len() {
```
pommicket, line 297: Maybe it's nicer to do something like
```
let file_name_bytes = file_name.iter().copied().take_while(|&byte| byte != 0).collect();
let mut file_name_str = String::from_utf8(file_name_bytes)
    .map_err(|_| Error::IO("bad UTF-8 in file name".into()))?;
```
 rather than this loop? (Then we handle non-ascii file names and catch invalid file names)

kernel/src/fs/vsfs/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+                    },
+                    direct_blocks: [0; VSFS_DIRECT_BLOCKS],
+                    indirect_block: 0,
+                    _padding: 0,
+                };
+
+                if inode_bitmap
+                    .is_allocated((i - VSFS_INODE_TABLE_BLOCK) * inode_ratio as u32 + k as u32)
+                {
+                    inode.mode = u32::from_le_bytes(
```
pommicket, line 162: If you do `use zerocopy::{FromBytes, FromZeroes}` and then add `FromBytes, FromZeroes` to the derives for `Timespec` and `Inode`, then you can replace all of this with `inode = Inode::read_from(&buffer[k * VSFS_INODE_SIZE..(k+1) * VSFS_INODE_SIZE]).unwrap()` :)

kernel/src/fs/vsfs/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+
+        // print out the data bitmap
+        //println!("Data bitmap: {:?}", vsfs.data_bitmap.bits);
+        println!("Data bitmap length: {}", vsfs.data_bitmap.bits.len());
+
+        // print out inode table
+        for i in 0..vsfs.inodes.len() {
+            // print out only allocated inodes in the inode bitmap
+            //println!("Inode {}: {:?}", i, vsfs.inodes[i]);
+            if (vsfs.inode_bitmap.is_allocated(i as u32)) {
```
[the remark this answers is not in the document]
minjaedavidpark: Fixed

kernel/src/fs/vsfs/mod.rs
minjaedavidpark: Wow, this is such a good tip! Thanks! (fixed)

pommicket (approved):
LGTM

PR #131, opened 2024-11-28
Add Pipe, Dup and Dup2 syscalls

1whatleytay:
This PR adds the `pipe`, `dup` and `dup2` syscalls. There are some things to note:

 - Specifically, when copying `Regular` OpenFiles, behaviour might be undefined. Seems like there is some handshake with the file system to indicate a file is open, but no handle is passed around.
 - Seems like fs would need some reference counting at some point, but since there is no concrete implementation, I just copy the contents of the OpenFile into the new handle.

Intuitively, it seems like most bugs with `dup` and `dup2` opening regular files should be pretty similar to the same opening a file via `open` twice. @pommicket if you'd like to double check this, I'd appreciate it.


kernel/src/fs/fs_manager.rs
```diff
@@ -836,6 +848,45 @@ impl RootFileSystem {
...
+        Ok((read_end.fd, write_end.fd))
+    }
+    pub fn dup(&mut self, pid: Pid, fd: ProcessFileDescriptor) -> Result<FileDescriptor> {
+        let open_file = self.open_files.get_mut(&fd).ok_or(Error::BadFd)?;
+
+        // Taking a look at the OpenFile type.
+        // Specifically, we have to be careful about Regular file types...
+        // The fs might need to know when to keep track of an inode.
+        // Currently, though, this doesn't look to be any less dangerous than opening a file twice.
+        // The fs doesn't keep a reference count, so someone in the future might need to tinker.
```
pommicket, line 871: Actually we do keep reference counts- see `FileSystemManager::open_file_count`
We definitely want to do `fs.inc_ref(inode)` for regular files here (you can get the inode from an fd using fstat currently which is a bit cumbersome but the mmap PR also adds an `inode_of` method)

pommicket (changes requested):
Overall looks really good!

kernel/src/fs/fs_manager.rs
1whatleytay: Thank you!! Let me do that. Wanted to bring that to light today.

1whatleytay: How's this?

pommicket (approved):
Looks great!

PR #134, opened 2024-11-28
Fix BitArray unable to clear single bits

### LINKED ISSUE #133: BitArray impl does not allow bits already set to be cleared
https://github.com/KidneyOS/KidneyOS/blob/903452b9bdd873ed017e056e13a90ae8c8866175/shared/src/bit_array.rs#L16-L18

The code shown above only allows us to set certain bits, but doesn't allow them to be cleared.

tonyhu-x:
Close #133


PrimeAce123 (approved):
Looks good

PR #136, opened 2024-11-28
Fix clippy lint

pommicket:
so that CI doesn't fail anymore


tonyhu-x:
Sorry, didn't see this and I went ahead and fixed it.

PR #142, opened 2024-11-29
Change GitHub action workflows

### LINKED ISSUE #139: Docs not being deployed in GitHub Action

tonyhu-x:
Close #139.

I also de-coupled the docs deployment workflow from the other checks. I didn't put a dependency relation since I don't think we need to necessarily wait for checks to pass before deploying docs. Plus, ideally the checks should always pass when we push to main, which is when the workflow is triggered.


.github/workflows/checks.yaml
```diff
@@ -0,0 +1,43 @@
+name: Run Checks
+on:
+  push
```
LanceLi1416, line 3: I would suggest ignoring the `docs` directory for this atcion.
```yaml
on:
  push
    paths-ignore:
      - 'docs/**'
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,49 @@
+name: Build and Deploy Docs
+on:
+  push:
+    branches:
+      - 'main'
```
LanceLi1416, line 5: And restrict this action to the `docs` directory.
```yaml
on:
  push:
    branches:
      - 'main'
    paths:
      - 'docs/**'
```

